PR #20, opened 2022-10-10
Add the tool multi-timestamp

### LINKED ISSUE #16: rdump timestamp field
The `runkey` plugin and other plugins have the field `ts`. The field `ts` can be used as timestamp field for all these plugins to analyze records with Kibana in chronological order.

```
target-query -f runkeys -q -t targets/MSEDGEWIN10_20220708124036.tar | rdump -w elastic://http://localhost:9200?index=timelines
```

image

But when using a plugin without the field `ts` these record cannot be shown in Kibana. For example `amcache` that does not have the field `ts` , because it uses other `datetime` fields like `created`.

```
target-query -f amcache -q -t targets/MSEDGEWIN10_20220708124036.tar | rdump -w elastic://http://localhost:9200?index=timelines
```

I don't know if it is supported, but is it possible that Dissect.target plugins that output records use the field `ts` that corresponds to the event creation timestamp of the parsed artefact? I am guessing Fox-IT uses Splunk which uses timestamp assignment which automatically assigns a timestamp based on what it finds in the raw even data. But this feels a bit like magic compared to using records with other adapters.

## References
[https://docs.splunk.com/Documentation/Splunk/latest/Data/HowSplunkextractstimestamps](https://docs.splunk.com/Documentation/Splunk/latest/Data/HowSplunkextractstimestamps)
[https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/os/windows/log/amcache.py#L17](https://github.com/fox-it/dissect.target/blob/main/dissect/target/plugins/os/windows/log/amcache.py#L17)
[https://www.elastic.co/guide/en/elasticsearch/reference/current/anonymous-access.html](https://www.elastic.co/guide/en/elasticsearch/reference/current/anonymous-access.html)
[https://www.elastic.co/guide/en/kibana/current/data-views.html](https://www.elastic.co/guide/en/kibana/current/data-views.html)

Zawadidone:
Fixes #16 

After merge add usage of multi-timestamp to docs https://github.com/fox-it/dissect-docs/blob/main/docs/source/tools/rdump.rst


yunzheng:
Hi Zawadidone,

I'm trying to determine if this script is useful enough to register it as it's own CLI tool, as the code example was just posted as an example so in my head it would fit better in `examples/` :)

But after reading #16 it seems that this is the only way to get proper timestamps (for some plugins) into Elastic/Kibana?

Zawadidone:
Yes Elasticsearch, Kibana and Timesketch all need a single `datetime` field for plugins that have `datetime` fields

yunzheng (approved):
I have refactored the multi timestamp code into it's own function so it can be properly tested.
Also added unit tests so CI doesn't fail.

Looks good for commit

yunzheng:
Merged!

Feel free to also work on the docs :)

Zawadidone:
Thanks I will create a PR for the documentation later on.

PR #22, opened 2022-11-01
Configure the minimum python version

pyrco:
The Python version dissect is developed and tested against is currently 3.9. It
is set as minimum version and not also as maximum version because newer
versions (like 3.10) might work and because it makes transitioning to newer
Python versions easier.

(DIS-1490)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,3 +9,6 @@ long_description_content_type = text/markdown
 license_files = LICENSE, COPYRIGHT
 classifiers =
   Programming Language :: Python :: 3
+
+[options]
+python_requires = ~=3.7


yunzheng:
I would prefer if the minimum version doesn't get bumped up to high because of some other project. `flow.record` in its current form is still perfectly compatible with Python 3.6. However, i can live if we target Python 3.7 instead so we can use nicer features.

So I suggest setting the minimum version to `3.7` instead of `3.9`.

pyrco:
> So I suggest setting the minimum version to `3.7` instead of `3.9`.

I'm fine with that, but note that the tests are only run for python 3.9. Though we plan to do linting on Python constructs which are not implemented in the minimum Python version supported.

PR #23, opened 2022-11-04
Don't output RecordDescriptor information in JSON output of `rdump`

yunzheng:
This disables outputting recorddescriptor information in the JSON output when using `--json` or `--jsonlines` as this is rarely what you want as an end user.

It's still possible to output recorddescriptor information if you use `jsonfile://<file>?descriptors=true` as the adapter.

(DIS-1637)

### DIFF
--- a/flow/record/adapter/jsonfile.py
+++ b/flow/record/adapter/jsonfile.py
@@ -9,7 +9,7 @@ from flow.record.fieldtypes import fieldtype_for_value
 __usage__ = """
 JSON adapter
 ---
-Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]&?descriptors=[DESCRIPTORS]
+Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]&descriptors=[DESCRIPTORS]
 Read usage: rdump jsonfile://[PATH]
 [PATH]: path to file. Leave empty or "-" to output to stdout
 [INDENT]: optional number of identation. Omit "indent" field value for jsonlines output
--- a/flow/record/adapter/jsonfile.py
+++ b/flow/record/adapter/jsonfile.py
@@ -9,22 +9,25 @@ from flow.record.fieldtypes import fieldtype_for_value
 __usage__ = """
 JSON adapter
 ---
-Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]
+Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]&?descriptors=[DESCRIPTORS]
 Read usage: rdump jsonfile://[PATH]
 [PATH]: path to file. Leave empty or "-" to output to stdout
 [INDENT]: optional number of identation. Omit "indent" field value for jsonlines output
+[DESCRIPTORS]: optional boolean. If false, don't output record descriptors (default: true)
 """
 
 
 class JsonfileWriter(AbstractWriter):
     fp = None
 
-    def __init__(self, path, indent=None, **kwargs):
+    def __init__(self, path, indent=None, descriptors=True, **kwargs):
+        self.descriptors = str(descriptors).lower() in ("true", "1")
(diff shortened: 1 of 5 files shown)


flow/record/adapter/jsonfile.py
```diff
@@ -9,22 +9,25 @@
 __usage__ = """
 JSON adapter
 ---
-Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]
+Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]&?descriptors=[DESCRIPTORS]
```
Schamper, line 12:
```suggestion
Write usage: rdump -w jsonfile://[PATH]?indent=[INDENT]&descriptors=[DESCRIPTORS]
```

tests/test_rdump.py
```diff
@@ -189,3 +189,58 @@ def test_rdump_json(tmp_path):
...
+            md5 = hashlib.md5(data).hexdigest()
+            sha1 = hashlib.sha1(data).hexdigest()
+            sha256 = hashlib.sha256(data).hexdigest()
+            writer.write(
+                TestRecord(
+                    count=i,
+                    foo="bar" * i,
+                    data=b"\x00\x01\x02\x03--" + data,
+                    ip="172.16.0.{}".format(i),
+                    subnet="192.168.{}.0/24".format(i),
```
Schamper, line 221:
```suggestion
                    ip=f"172.16.0.{i}",
                    subnet=f"192.168.{i}.0/24",
```

tests/test_rdump.py
```diff
@@ -189,3 +189,58 @@ def test_rdump_json(tmp_path):
...
+    for i, line in enumerate(stdout.splitlines()):
+        json_dict = json.loads(line)
+        assert json_dict
+        assert "_type" not in json_dict
+        assert "_recorddescriptor" not in json_dict
+        assert "_source" in json_dict
+        data = str(i).encode()
+        assert json_dict["data"] == base64.b64encode(b"\x00\x01\x02\x03--" + data).decode()
+        assert json_dict["ip"] == "172.16.0.{}".format(i)
+        assert json_dict["subnet"] == "192.168.{}.0/24".format(i)
```
Schamper, line 243:
```suggestion
        assert json_dict["ip"] == f"172.16.0.{i}"
        assert json_dict["subnet"] == f"192.168.{i}.0/24"
```

Schamper (changes requested):

Schamper (approved):

Zawadidone:
@yunzheng should usage of `jsonfile://<file>` output the record descriptor? I would have not expected that because the parameter `?descriptors=True` is not specified.

```bash
target-query -t targets/* -f appinit -q | rdump -w jsonfile://test1.jsonl
[reading from stdin]
root@processing-anti:/mnt/processing# head -n 1 test1.jsonl 
{"_type": "recorddescriptor", "_data": ["filesystem/registry/appinit", [["string", "hostname"], ["string", "domain"], ["datetime", "ts"], ["uri", "path"], ["string", "regf_hive_path"], ["string", "regf_key_path"], ["string", "username"], ["string", "user_id"], ["string", "user_home"]]]}
```

Schamper:
@Zawadidone looks like `True` is the default value, so specifying `False` will actually get the behavior you're looking for.

yunzheng:
@Zawadidone I left the default to `True` for the record adapter itself, as I was not sure if it would introduce breaking changes to other plugins that might use the JSON adapter. It's only set to `False` in the `rdump` tool when using the `--json` or `--jsonlines` flags, you could use output redirection instead of using the `jsonfile://` uri:

```shell
target-query -t targets/* -f appinit -q | rdump -J > test1.jsonl
```

Let us know if you think this behaviour should be changed.

Zawadidone:
It works great. This makes it possible to dynamicly ingest rdump JSON files to Elasticsearch/Timesketch.

PR #30, opened 2022-11-29
Fix `fieldtypes.path` initializer and added extra tests

### LINKED ISSUE #29: Incorrect value when initializing `path` fieldtype with `PureWindowsPath`
For some Windows path values the `windows_path` field type is incorrectly initialized, for example:

```python
>>> import pathlib
>>> from flow.record import fieldtypes
>>> fieldtypes.path(pathlib.PureWindowsPath("/c:/Windows/System32/drivers/null.sys")
windows_path('c:Windows/System32/drivers/null.sys')
```

yunzheng:
Fixes #29

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -521,7 +521,7 @@ def test_windows_path_regression(path_initializer):
         [
             ("path", "path"),
         ],
-    )    
+    )
     r = TestRecord(path=path_initializer("/c:/Windows/System32/drivers/null.sys"))
     assert str(r.path) == "\\c:\\Windows\\System32\\drivers\\null.sys"
     assert repr(r.path) == "windows_path('/c:/Windows/System32/drivers/null.sys')"


flow/record/fieldtypes/__init__.py
```diff
@@ -545,8 +545,10 @@ def __new__(cls, *args):
                 path_ = args[0]
                 if isinstance(path_, pathlib.PureWindowsPath):
                     cls = windows_path
+                    args = [path_.as_posix()]
```
pyrco, line 548: Interesting issue. This feels like a problem with `pathlib.PureWindowsPath` if initializing it with a `PureWindowsPath` constructs it as a slightly different `PureWindowsPath`.

pyrco (approved):

flow/record/fieldtypes/__init__.py
yunzheng: Yes, think it's also an issue in Python.

PR #35, opened 2022-12-08
Fix `rdump --count` and `rdump --list` counting

yunzheng:


yunzheng:
The count was off by one in `--list` output

Schamper (approved):

PR #37, opened 2023-01-03
Fix flow.record datetime for timezones

JSCU-CNI:
Add support for parsing datetime strings with `%z` timezone information.


Schamper:
Can you add a unit test that includes timezone info?

yunzheng:
Note that `flow.record` packer currently does not support serializing timezone aware datetime objects, it's always serialized and deserialized as UTC.

yunzheng:
Looks good to me.

@Schamper if you have no further remarks you can merge it.

yunzheng (approved):

PR #41, opened 2023-01-10
Add `--multi-timestamp` to rdump

### LINKED ISSUE #40: Rdump multi-timestamp functionality
I would like to speed up to processing of `rmult-timestamp` by allowing it to read JSON files or JSON input from standard input.

As a result the following functionality should be added:
* Use `target-query [...] -j` to create JSON files or output JSON (already supported)
* Use `rmulti-timestamp` to read JSON from standard input or from a file (currently not supported) and output record format (already supported) or JSON format (currently not supported)

I have a few questions to better understand how this could be added to the project:
* Should this implementation be the same as how `rdump` works with reading and outputing?
* What can of functionality is new and cannot be re-used from current code?

Zawadidone:
Closes #40
Adjust documentation on `rdump` multi-timestamp usage [https://github.com/fox-it/dissect-docs/commit/34fd3c1319385ac923a3a48db435fee01c4b1831](https://github.com/fox-it/dissect-docs/commit/34fd3c1319385ac923a3a48db435fee01c4b1831)


flow/record/tools/rdump.py
```diff
@@ -24,8 +23,9 @@
 
 try:
     # Python 2
-    import urlparse
     from urllib import urlencode
+
+    import urlparse
 except ImportError:
     # Python 3
     import urllib.parse as urlparse
```
yunzheng, line 31: When touching Python 2 code, just remove it as we don't support it anymore.

yunzheng (changes requested):

flow/record/tools/rdump.py
Zawadidone: Code is removed

flow/record/tools/rdump.py
```diff
@@ -3,9 +3,11 @@
 
 import logging
 import sys
+import urllib.parse as urlparse
 from importlib import import_module
 from pathlib import Path
 from textwrap import indent
+from urllib.parse import urlencode
```
yunzheng, line 10: Can you change the urllib imports to:

```python
from urllib.parse import urlencode, urlparse
```

And then change the `urlparse.urlparse` code to just `urlparse`.

yunzheng (changes requested):
Small Python2 compat thing that can be refactored out. Then it looks good for merge.

flow/record/tools/rdump.py
Zawadidone: Adjusted

yunzheng (approved):

PR #43, opened 2023-01-12
Use newline='' for CsvfileWriter

### LINKED ISSUE #42: rdump csv output has empty CR LF lines
Both `rdump <input file> -w csvfile://<output file>` and `rdump <input file> -C > <output file>`  create csv file in which every line ends with CR followed by CR LF. As CR ends line, result is that every second line of csv file is empty line containing only CR LF.

I am using Python 3.9.5 and Windows 10.

yunzheng:
Fixes #42

This commit also adds the `lineterminator` option to the csvfile adapter
It defaults to `\r\n`, which is also the Python default.

If you want to output csv files with only `\n`, you can do so using:

    csvfile://?lineterminator=\n

### DIFF
--- a/flow/record/adapter/csvfile.py
+++ b/flow/record/adapter/csvfile.py
@@ -9,7 +9,7 @@ from flow.record.adapter import AbstractWriter
 __usage__ = """
 Comma-separated values (CSV) adapter
 ---
-Write usage: rdump -w csvfile://[PATH]&lineterminator=[TERMINATOR]
+Write usage: rdump -w csvfile://[PATH]?lineterminator=[TERMINATOR]
 Read usage: rdump csvfile://[PATH]
 [PATH]: path to file. Leave empty or "-" to output to stdout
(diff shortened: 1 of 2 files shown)


flow/record/adapter/csvfile.py
```diff
@@ -3,27 +3,30 @@
...
 
-from flow.record import open_path
 from flow.record.utils import is_stdout
 from flow.record.adapter import AbstractWriter
 
 __usage__ = """
 Comma-separated values (CSV) adapter
 ---
-Write usage: rdump -w csvfile://[PATH]
+Write usage: rdump -w csvfile://[PATH]&lineterminator=[TERMINATOR]
```
Schamper, line 12:
```suggestion
Write usage: rdump -w csvfile://[PATH]?lineterminator=[TERMINATOR]
```

Schamper (changes requested):

Schamper (approved):

PR #45, opened 2023-01-20
Allow serialization and deserialization of `surrogateescape` strings

cecinestpasunepipe:
Make flow record more robust against invalid UTF-8 byte sequences like UTF-16 surrogate-like sequences

(DIS-1735)

### DIFF
--- a/flow/record/packer.py
+++ b/flow/record/packer.py
@@ -8,7 +8,7 @@ from .base import Record, FieldType, RecordDescriptor, GroupedRecord, RESERVED_F
 from .utils import EventHandler, to_str
 
 # Override defaults for msgpack packb/unpackb
-packb = functools.partial(msgpack.packb, use_bin_type=True)
+packb = functools.partial(msgpack.packb, use_bin_type=True, unicode_errors="backslashreplace")
 unpackb = functools.partial(msgpack.unpackb, raw=False)
 
 RECORD_PACK_EXT_TYPE = 0xE
--- a/flow/record/packer.py
+++ b/flow/record/packer.py
@@ -8,8 +8,8 @@ from .base import Record, FieldType, RecordDescriptor, GroupedRecord, RESERVED_F
 from .utils import EventHandler, to_str
 
 # Override defaults for msgpack packb/unpackb
-packb = functools.partial(msgpack.packb, use_bin_type=True, unicode_errors="backslashreplace")
-unpackb = functools.partial(msgpack.unpackb, raw=False)
+packb = functools.partial(msgpack.packb, use_bin_type=True, unicode_errors="surrogateescape")
+unpackb = functools.partial(msgpack.unpackb, raw=False, unicode_errors="surrogateescape")
 
 RECORD_PACK_EXT_TYPE = 0xE
 
--- a/tests/test_fieldtypes.py
+++ b/tests/test_fieldtypes.py
@@ -757,5 +757,44 @@ def test_format_hex(spec, value, expected):
     assert format_str.format(record.value) == expected
 
 
+@pytest.mark.parametrize(
+    "filename",
+    [
+        "test.records",
+        "test.records.gz",
+        "test.records.json",
+    ],
+)
+@pytest.mark.parametrize(
+    "str_bytes,unicode_errors,expected_str",
+    [
+        (b"hello \xa7 world", "surrogateescape", "hello \udca7 world"),
+        (b"hello \xa7 world", "backslashreplace", "hello \\xa7 world"),
+        (b"hello \xa7 world", "replace", "hello \ufffd world"),
+        (b"hello \xa7 world", "ignore", "hello  world"),
+    ],
+)
+def test_string_serialization(tmp_path, filename, str_bytes, unicode_errors, expected_str):
+    TestRecord = RecordDescriptor(
+        "test/record",
+        [
+            ("string", "str_value"),
+        ],
+    )
+
+    str_value = str_bytes.decode("utf-8", errors=unicode_errors)
+    record = TestRecord(str_value=str_value)
+    assert str_value == expected_str
+    assert record.str_value == expected_str
+
+    with RecordWriter(tmp_path / filename) as writer:
+        writer.write(record)
+
+    with RecordReader(tmp_path / filename) as reader:
+        record = next(iter(reader))
+        assert str(record.str_value) == expected_str
+        assert record.str_value == expected_str
+
+
 if __name__ == "__main__":
     __import__("standalone_test").main(globals())
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -605,5 +605,28 @@ def test_datetime_as_fieldname():
     TestRecord()
 
 
+def test_string_surrogateescape_serialization(tmp_path):
+    TestRecord = RecordDescriptor(
+        "test/record",
+        [
+            ("string", "str_value"),
+        ],
+    )
+
+    str_value = b"hello \xa7 world".decode(errors="surrogateescape")
+    record = TestRecord(str_value=str_value)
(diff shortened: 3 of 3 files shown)


yunzheng:
do you have a test case? Also the documention of msgpack states to not use `unicode_errors`: https://msgpack-python.readthedocs.io/en/latest/api.html#msgpack.Packer

cecinestpasunepipe:
> do you have a test case? Also the documention of msgpack states to not use `unicode_errors`: https://msgpack-python.readthedocs.io/en/latest/api.html#msgpack.Packer

Testcase: this issue is a consequence of: https://github.com/fox-it/dissect.target/pull/137
If you use surrogatereplace you'll get an error. Now, I am not advocating the use of surrogates but they work fine to demonstrate the effect of invalid byte sequences.

Sure, the docs at msgpack recommend not to use `unicode_errors` - but that is for the regular use case. We on the other hand are interested in preserving as much data as possible. In a traditional context, throwing exceptions would be the best way to handle invalid sequences, because it might point to an encoding issue or security problem deeper in the code. However, in our case, we are interested (I presume) in the invalid sequence itself because it might reveal an attack vector for instance. Also, in our case, we know the "data is dirty" so there is no need for errors, because such anomalies in the data are expected as it might come from a compromised source. In our case it does not necessarily indicate a problem with our own code, so it does not make sense to throw errors or hide the data. Point is, we are just an atypical user of msgpack and in our situation the initial warning does not apply (they also state in the docs that this feature is meant for people who know what they are doing = this is us).

Hopefully I have explained the rationale behind this change from my perspective clear enough? Please tell me if I overlook something here... always possible of course...

flow/record/packer.py
```diff
@@ -8,7 +8,7 @@
 from .utils import EventHandler, to_str
 
 # Override defaults for msgpack packb/unpackb
-packb = functools.partial(msgpack.packb, use_bin_type=True)
+packb = functools.partial(msgpack.packb, use_bin_type=True, unicode_errors="backslashreplace")
```
yunzheng, line 11: Thanks for the explanation and small example case, however this change shows you use `unicode_errors="backslashreplace"`  did you mean to use `unicode_errors="surrogateescape"`?

yunzheng (changes requested):

flow/record/packer.py
cecinestpasunepipe: No, I meant `backslashreplace`. This will reveal the bytes as-is for inspection. Per PEP-383 (https://peps.python.org/pep-0383/) the surrogateescape-handler allows developers to smuggle bytes across systems to facilitate non-UTF-8 integration, this is not the intention of Dissect I assume. In Dissect a clear byte-as-is description of invalid sequences would be preferable I guess.

yunzheng: Then, why not directly use `backslashreplace` as the input (taken from your example):

```python
import sys
from flow.record import RecordDescriptor, RecordStreamWriter

rd = RecordDescriptor("desc", [("string", "data")])
rs = RecordStreamWriter(sys.stdout.buffer)
d = b"Hello \xa7 World".decode("utf-8", errors="backslashreplace")
rs.write(rd(data=d))
```

Output:

```shell
(dissect) user@mbp ~> python3 /tmp/test.py | rdump -L
[reading from stdin]
--[ RECORD 1 ]--
           data = Hello \xa7 World
        _source = None
_classification = None
     _generated = 2023-01-23 15:27:47.592665
       _version = 1
```

cecinestpasunepipe: That is indeed exactly what I am going to do in that PR. However, I thought it would be nice to have flow.record not raise an exception as well, there can be sources that do not perform this transformation right?

cecinestpasunepipe: Well, the problem is that you can never tell what these bytes are... in theory they could be anything. Unless you want to limit the input of flow.record to "known tools", but then we need to communicate this to the users. Also, I thought flow.record was more like a general tool that could also accept data streams from other sources, but maybe this is not the case?

cecinestpasunepipe: I was under the assumption that we cannot be sure about the input, although we prescribe UTF-8. But maybe we just want to make this a requirement (I do not see this in the docs... but maybe I am overlooking something)

yunzheng: The thing is that the fieldtype is a `str` so if from a user perspective when you are parsing bytes and want to save it as a `str` the user needs to do a `.decode()` on it. Or else you can just store it in a `bytes` fieldtype correct?

I think it’s the user’s responsibility to choose the unicode error handling. if you assign bytes to a `flow.record` string field it will currently just do `.decode()` without any error handling. We can decide to force this to “backslashreplace” for example. If this is the intent then it needs to be changed somewhere else and not in the packer code.

if we want to allow “surrogateescaped” unicode strings we need to patch the packer and unpacker. So that’s why i was asking what the intent was. It’s also possible i’m overlooking something, if that is the case let me know.

yunzheng: To recap, the msgpack packer doesn't allow 
serialization and deserialization of “surrogateescaped” strings by default. Any other python unicode str works fine from what i can tell. 

if we want to serialize bytes as a string and we want to choose a default error handling for the user (so that it always works), we need to change this in the string fieldtype code.

cecinestpasunepipe: I can understand the reasoning that flow.record is not responsible for invalid UTF-8 byte sequences. That is fine by me. Might be a good idea to mention the valid UTF-8 requirement both in the rdump manual and the flow.record manual though. Otherwise I fear that this might become a recurring topic.

yunzheng: Do you have an example of an invalid UTF-8 `str` that is not `surrogateescape`'ed?

Schamper:
@cecinestpasunepipe shouldn't this be renamed (and changed) to allow surrogate escape'd strings to be serialized by flow.record? So setting `unicode_errors="surrogateescape"`.

Practical example: all filesystem implementations use `errors="surrogateescape"` because that's the standard expected behaviour of decoding file names. If we try to serialize such a file name currently with Dissect this will result in an error. Setting `unicode_errors="surrogateescape"` for both serializing and deserializing will result in the exact same output as input, and thus in the correct file name within a record.

yunzheng:
Good to know that we are using `surrogateescape` strings in some dissect projects, so `flow.record` should definitely allow serializing and deserializing of this.

I'm reopening the PR and rename the title.

yunzheng:
I've added a regression test and some additional unit tests, feel free to review @Schamper @cecinestpasunepipe

yunzheng (approved):

cecinestpasunepipe:
I would recommend the printable backslashreplace approach, simply because this makes inspection easier. Delegating responsibility to the input channel would be my second choice. Surrogates seem the worst of both worlds but there might be practical advantages that I do not see or understand.

yunzheng:
> I would recommend the printable backslashreplace approach, simply because this makes inspection easier. Delegating responsibility to the input channel would be my second choice. Surrogates seem the worst of both worlds but there might be practical advantages that I do not see or understand.

I think if the user chooses to create a `surrogateescape` string from bytes, and you use this within `flow.record`, that it should be possible to do so, and that if you serialize it and deserialize it, it results into the same exact `str` object. This is what the current commit provides.

And yes, the responsibility is primary the user. `flow.record` just facilitates. If we used `backslashreplace` in msgpack and someone wants to serialize a `surrogateescape` string, you cannot get back the same original `str` object as it would then have applied a `backslashreplace` on the `surrogate` string. I don't think this is what we want.

If the user chooses to decode a byte string using `backslashreplace` and wants to serialize this in `flow.record`, it works just fine already without any changes.

cecinestpasunepipe:
LGTM

Issue #46, opened 2023-01-23
rdump splitted JSON files leads to key error


Zawadidone:
Create splitted JSON files

```bash
target-query -f mft -t MSEDGEWIN10.tar --limit 10 -j > mft.json

# Split single file in two files
split -dl 6 mft.json
mv x00 x00.json
mv x01 x01.json
```

Dump records. The first file outputed can be read with `rdump`, but all files after that result in a JSON key error.

```bash
 rdump x00.json 
<filesystem/ntfs/mft/std hostname='MSEDGEWIN10' domain=None creation_time=2019-03-19 21:52:25.169411 last_modification_time=2019-03-19 21:52:25.169411 last_change_time=2019-03-19 21:52:25.169411 last_access_time=2019-03-19 21:52:25.169411 segment=0 path='c:/$MFT' owner='S-1-5-18' filesize=0.12 GB resident=False inuse=True volume_uuid=None>
[...]

rdump x01.json 
2023-01-23 15:31:32,042 WARNING Exception in <flow.record.adapter.jsonfile.JsonfileReader object at 0x10a0dec50> for 'x01.json': KeyError(('filesystem/ntfs/mft/std', 623539933)) -- skipping to next reader
```

Zawadidone:
The variable `self.descriptors` is empty, because the JSON file `x01.json` does not contain a record descriptor record at line 1. Because of that line 83 leads to a key error [https://github.com/fox-it/flow.record/blob/60f9a731681eb0550c404003c4f9151d439dc9b1/flow/record/jsonpacker.py#L83](https://github.com/fox-it/flow.record/blob/60f9a731681eb0550c404003c4f9151d439dc9b1/flow/record/jsonpacker.py#L83)

yunzheng:
Yes the descriptors are only dumped at the beginning of the json file (as a single entry/line). I think the best way to have proper support is to add a `--split` option to `rdump`.

Zawadidone:
So that feature would do the following:

* `rdump --split 6 mft.json`
* Read the record descriptor
* Read 6 lines and create file `mft<suffix>.json` that starts with the record descriptor and the 6 lines (repeat until end of file).

yunzheng:
No, instead of reading I would mean you need to dump the records as splitted, eg: `rdump --split 6 -w mft.json` or `rdump --split 100 -w my.records.gz`.

It would then dump (the postfix numbering is up for discussion):

 * mft.1.json or my.records.1.gz
[...]

PR #47, opened 2023-01-25
Add `--split` option to `rdump`

yunzheng:
This implements #46, which allows you to split records over multiple smaller output files.

Example usage:

```shell
$ target-query -f mft -t MSEDGEWIN10.tar --limit 10 | rdump --split 6 -w mft.json
$ wc -l mft*.json
       8 mft.00.json
       6 mft.01.json
      14 total

$ rdump mft.00.json -l | grep Processed
Processed 6 records

$ rdump mft.01.json -l | grep Processed
Processed 4 records
```

### DIFF
--- a/flow/record/adapter/split.py
+++ b/flow/record/adapter/split.py
@@ -1,0 +1,75 @@
+from pathlib import Path
+
+from urllib.parse import urlparse
+
+from flow.record.adapter import AbstractWriter
+from flow.record.base import RecordWriter
+
+DEFAULT_RECORD_COUNT = 1000
+DEFAULT_SUFFIX_LENGTH = 2
+
+__usage__ = f"""
+Record split adapter, splits records into multiple destination files (writer only)
+---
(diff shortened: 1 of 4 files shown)


flow/record/adapter/split.py
```diff
@@ -0,0 +1,75 @@
...
+        if self.is_stdout:
+            return self.path
+
+        path = self.path
+        scheme = ""
+        sep = ""
+        if "://" in path:
+            scheme, sep, path = path.partition("://")
+
+        suffix = f"{self.file_count}".rjust(self.suffix_length, "0")
```
Schamper, line 48:
```suggestion
        suffix = str(self.file_count).rjust(self.suffix_length, "0")
```
? Or not, maybe feels a bit simpler.

Schamper (changes requested):

flow/record/adapter/split.py
yunzheng: yes it is, thx

Schamper (approved):

PR #50, opened 2023-01-26
UnknownRecordDescriptorIdentifier Exception

cecinestpasunepipe:
(DIS-1603)

### DIFF
--- a/flow/record/packer.py
+++ b/flow/record/packer.py
@@ -1,10 +1,19 @@
-import warnings
 import datetime
-import msgpack
 import functools
+import warnings
+
+import msgpack
 
 from . import fieldtypes
-from .base import Record, FieldType, RecordDescriptor, GroupedRecord, RESERVED_FIELDS, RECORD_VERSION
+from .base import (
+    RECORD_VERSION,
+    RESERVED_FIELDS,
+    FieldType,
+    GroupedRecord,
+    Record,
+    RecordDescriptor,
+)
+from .exceptions import RecordDescriptorNotFound
 from .utils import EventHandler, to_str
 
 # Override defaults for msgpack packb/unpackb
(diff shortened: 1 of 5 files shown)


flow/record/packer.py
```diff
@@ -28,6 +28,10 @@ def identifier_to_str(identifier):
         return to_str(identifier)
 
 
+class UnknownRecordDescriptorIdentifier(Exception):
```
yunzheng, line 31:
```suggestion
class RecordDescriptorNotFound(Exception):
```

I think this Exception name is a bit more readable.

flow/record/packer.py
```diff
@@ -107,6 +111,10 @@ def unpack_obj(self, t, data):
         if subtype == RECORD_PACK_TYPE_RECORD:
             identifier, values = value
             identifier = identifier_to_str(identifier)
+
+            if identifier not in self.descriptors:
+                raise UnknownRecordDescriptorIdentifier(f"No record descriptor found for: {identifier}")
```
yunzheng, line 116:
```suggestion
                raise RecordDescriptorNotFound(f"No record descriptor found for: {identifier}")
```

yunzheng (changes requested):
Can you also add a unit test?

Schamper:
Should this also be added to `jsonpacker.py`?

flow/record/packer.py
```diff
@@ -107,6 +111,10 @@ def unpack_obj(self, t, data):
         if subtype == RECORD_PACK_TYPE_RECORD:
             identifier, values = value
             identifier = identifier_to_str(identifier)
+
+            if identifier not in self.descriptors:
+                raise RecordDescriptorNotFound(f"No RecordDescriptor found for: {identifier}")
+
             desc = self.descriptors[identifier]
```
Schamper, line 118:
```suggestion
            desc = self.descriptors.get(identifier)
            if not desc:
                raise RecordDescriptorNotFound(f"No RecordDescriptor found for: {identifier}")
```
This might be more efficient? As this is a very hot code path.

yunzheng (approved):
I've updated the tests to use `pytest.raises` and also make use of two different packer instances to clarify the intent of the test.

LGTM

PR #51, opened 2023-01-26
Add caching of RecordDescriptors to `extend_record()`

### LINKED ISSUE #48: Cache the creation of RecordDescriptors in `extend_record()`
The method `extend_record()` generates a new RecordDescriptor every time, which is an expensive task. The creation of the RecordDescriptors should be cached.

This will directly improve the speed where `extend_record()` is used, such as the `--multi-timestamp` option in `rdump`.

yunzheng:
Fixes #48

This greatly improves the speed of `--multi-timestamp` or other cases where `extend_record()` is called often.

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -756,6 +756,30 @@ def fieldtype(clspath):
     return t
 
 
+@functools.lru_cache(maxsize=4069)
+def merge_record_descriptors(descriptors, replace=False, name=None):
+    """Create a newly merged RecordDescriptor from a list of RecordDescriptors.
+    This function uses a cache to avoid creating the same descriptor multiple times.
+
+    Duplicate fields are ignored in `descriptors` unless `replace=True`.
+
+    Args:
+        descriptors (Tuple[RecordDescriptor]): Tuple of RecordDescriptors we use for extending/replacing.
+        replace (bool): if `True`, it will replace existing field names, eg: last descriptor always wins.
+        name (str): rename the RecordDescriptor name to `name`. Otherwise, use name from first descriptor.
+
+    Returns:
+        RecordDescriptor: Merged RecordDescriptor
+    """
+    field_map = collections.OrderedDict()
+    for desc in descriptors:
+        for (ftype, fname) in desc.get_field_tuples():
+            if not replace and fname in field_map:
+                continue
+            field_map[fname] = ftype
+    return RecordDescriptor(name or descriptors[0].name, zip(field_map.values(), field_map.keys()))
+
+
 def extend_record(record, other_records, replace=False, name=None):
     """Extend `record` with fields and values from `other_records`.
 
@@ -768,20 +792,17 @@ def extend_record(record, other_records, replace=False, name=None):
             in `record` from fields and values from `other_records`. Last record always wins.
         name (str): rename the RecordDescriptor name to `name`. Otherwise, use name from
             initial `record`.
+
+    Returns:
+        Record: Extended Record
     """
-    field_map = collections.OrderedDict((fname, ftype) for (ftype, fname) in record._desc.get_field_tuples())
-    record_maps = [record._asdict()]
-    for other in other_records:
-        for (ftype, fname) in other._desc.get_field_tuples():
-            if not replace and fname in field_map:
-                continue
-            field_map[fname] = ftype
-        record_maps.append(other._asdict())
-    field_tuples = [(ftype, fname) for (fname, ftype) in field_map.items()]
-    ExtendedRecord = RecordDescriptor(name or record._desc.name, field_tuples)
+    records = (record, *other_records)
+    descriptors = tuple(rec._desc for rec in records)
+    ExtendedRecord = merge_record_descriptors(descriptors, replace, name)
+    kv_maps = tuple(rec._asdict() for rec in records)
     if replace:
-        record_maps = record_maps[::-1]
-    return ExtendedRecord.init_from_dict(collections.ChainMap(*record_maps))
+        kv_maps = kv_maps[::-1]
+    return ExtendedRecord.init_from_dict(collections.ChainMap(*kv_maps))
 
 
 class DynamicFieldtypeModule:
--- a/tests/test_multi_timestamp.py
+++ b/tests/test_multi_timestamp.py
@@ -2,6 +2,7 @@ import datetime
 
 from flow.record import RecordDescriptor
 from flow.record import iter_timestamped_records
+from flow.record.base import merge_record_descriptors
 
 
 def test_multi_timestamp():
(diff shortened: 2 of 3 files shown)


flow/record/base.py
```diff
@@ -756,6 +756,30 @@ def fieldtype(clspath):
     return t
 
 
+@functools.lru_cache(maxsize=4069)
+def merge_record_descriptors(descriptors, replace=False, name=None):
```
Schamper, line 760: Maybe start adding type hints to new pieces of code? Don't need to do `base.py` in its entirety but new or modified functions?

flow/record/base.py
```diff
@@ -756,6 +756,30 @@ def fieldtype(clspath):
...
+    """Create a newly merged RecordDescriptor from a list of RecordDescriptors.
+    This function uses a cache to avoid creating the same descriptor multiple times.
+
+    Duplicate fields are ignored in `descriptors` unless `replace=True`.
+
+    Args:
+        descriptors (Tuple[RecordDescriptor]): Tuple of RecordDescriptors we use for extending/replacing.
```
Schamper, line 767:
```suggestion
        descriptors (Tuple[RecordDescriptor]): Tuple of RecordDescriptors to use for extending/replacing.
```

Schamper: Also would a type of `Sequence` or `Iterable` not be more accurate?

flow/record/base.py
```diff
@@ -756,6 +756,30 @@ def fieldtype(clspath):
...
+        replace (bool): if `True`, it will replace existing field names, eg: last descriptor always wins.
```
Schamper, line 768:
```suggestion
        replace (bool): if ``True``, it will replace existing field names. Last descriptor always wins.
```
It's not really an example if it's a rule.

flow/record/base.py
```diff
@@ -756,6 +756,30 @@ def fieldtype(clspath):
     return t
 
 
+@functools.lru_cache(maxsize=4069)
+def merge_record_descriptors(descriptors, replace=False, name=None):
+    """Create a newly merged RecordDescriptor from a list of RecordDescriptors.
+    This function uses a cache to avoid creating the same descriptor multiple times.
+
+    Duplicate fields are ignored in `descriptors` unless `replace=True`.
```
Schamper, line 764:
```suggestion
    Duplicate fields are ignored in ``descriptors`` unless ``replace=True``.
```

flow/record/base.py
```diff
@@ -756,6 +756,30 @@ def fieldtype(clspath):
...
+
+    Args:
+        descriptors (Tuple[RecordDescriptor]): Tuple of RecordDescriptors we use for extending/replacing.
+        replace (bool): if `True`, it will replace existing field names, eg: last descriptor always wins.
+        name (str): rename the RecordDescriptor name to `name`. Otherwise, use name from first descriptor.
```
Schamper, line 769:
```suggestion
        name (str): rename the RecordDescriptor name to ``name``. Otherwise, use name from first descriptor.
```

flow/record/base.py
```diff
@@ -756,6 +756,30 @@ def fieldtype(clspath):
...
+    Returns:
+        RecordDescriptor: Merged RecordDescriptor
+    """
+    field_map = collections.OrderedDict()
+    for desc in descriptors:
+        for (ftype, fname) in desc.get_field_tuples():
+            if not replace and fname in field_map:
+                continue
+            field_map[fname] = ftype
+    return RecordDescriptor(name or descriptors[0].name, zip(field_map.values(), field_map.keys()))
```
Schamper, line 780: This will error if `descriptors` is empty, maybe put a guard up for that.

Schamper (changes requested):

flow/record/base.py
yunzheng: In this case it has to be a Tuple, otherwise it's not hashable for lru_cache.

flow/record/base.py
yunzheng: Yes good one, see new commit. I'm also thinking about deprecating some initialisation code of `RecordDescriptor`. But saving that for another PR.

flow/record/base.py
yunzheng: Refactored it a bit, it can still be `None` which should not be allowed, but this now raises an Exception in RecordDescriptor's __init__().

When #50 is merged, i will rebase and move `RecordDescriptorError` to `flow/record/exceptions.py`

yunzheng:
There are also caching opportunities in RecordDescriptor itself, was thinking of doing it in this same PR, but it's probably better to split it or make it part of a bigger RecordDescriptor refactor. Currently RecordDescriptor has two other methods it can be used for initialisation but we should deprecate them as this is mostly legacy code and not used afaik.

flow/record/base.py
```diff
@@ -347,6 +347,9 @@ class RecordDescriptor:
     _desc_hash = None
 
     def __init__(self, name, fields=None):
+        if not name:
+            raise RecordDescriptorError(f"Record name is required, got {name!r}")
```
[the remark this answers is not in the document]
yunzheng: Yes, I can also leave it to just "Record name is required".

flow/record/base.py
```diff
@@ -756,32 +759,55 @@ def fieldtype(clspath):
...
             field_map[fname] = ftype
-        record_maps.append(other._asdict())
-    field_tuples = [(ftype, fname) for (fname, ftype) in field_map.items()]
-    ExtendedRecord = RecordDescriptor(name or record._desc.name, field_tuples)
+    if name is None and descriptors:
+        name = descriptors[0].name
+    return RecordDescriptor(name, zip(field_map.values(), field_map.keys()))
+
+
+def extend_record(record, other_records, replace=False, name=None):
```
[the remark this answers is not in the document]
yunzheng: done

flow/record/base.py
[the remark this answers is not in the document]
yunzheng: Simplified in new commit

flow/record/base.py
```diff
@@ -756,32 +757,59 @@ def fieldtype(clspath):
...
-        record_maps.append(other._asdict())
-    field_tuples = [(ftype, fname) for (fname, ftype) in field_map.items()]
-    ExtendedRecord = RecordDescriptor(name or record._desc.name, field_tuples)
+    if name is None and descriptors:
+        name = descriptors[0].name
+    return RecordDescriptor(name, zip(field_map.values(), field_map.keys()))
+
+
+def extend_record(
+    record: Record, other_records: List[Record], replace: Optional[bool] = False, name: Optional[str] = None
```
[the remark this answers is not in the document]
yunzheng: whoops, nice catchu

Schamper (approved):

PR #53, opened 2023-01-31
Add csv file reading support

### LINKED ISSUE #52: Implement `CsvfileReader` to `csvfile.py` adapter
Currently there is only writing support for CSV files, but in some cases it would be nice to also read CSV files and output them as records.

This does not have to be a fully featured unpacker that supports field types.

yunzheng:
Fixes #52


flow/record/adapter/csvfile.py
```diff
@@ -52,3 +54,28 @@ def close(self):
...
         self.fp = None
+
+
+class CsvfileReader(AbstractReader):
+    fp = None
+
+    def __init__(self, path, selector=None, **kwargs):
+        self.selector = make_selector(selector)
+        if path in (None, "", "-"):
+            self.fp = sys.stdout
```
Schamper, line 65:
```suggestion
            self.fp = sys.stdin
```
I assume?

flow/record/adapter/csvfile.py
```diff
@@ -52,3 +54,28 @@ def close(self):
...
+        self.reader = csv.reader(self.fp)
+
+    def close(self):
+        if self.fp:
+            self.fp.close()
+        self.fp = None
+
+    def __iter__(self):
+        row = next(self.reader)
+        desc = RecordDescriptor("csv/reader", [("string", col) for col in row])
```
Schamper, line 77: Maybe nice feature to support headerless CSV files and specify the columns through an argument of the adapter?

Schamper (changes requested):

flow/record/adapter/csvfile.py
yunzheng: oh oops, nice catch.

flow/record/adapter/csvfile.py
yunzheng: hmm yes, was thinking this as well. Will look into it.

yunzheng: Adding it to the adapter should be no problem, would you also expect this to be easy to do via `rdump` as well, for example via a new CLI argument, or would specifying a `csvfile://stuff.csv?fields=a,b,c` be sufficient?

Schamper: Yeah I meant the latter method, supplying through the adapter query arguments should be sufficient.

Schamper (approved):

PR #56, opened 2023-02-02
Refactor and cleanup RecordDescriptor class

yunzheng:
This commit adds the following improvements:

 * Move Record code generation to `_generate_record_class` which is also cached
 * Add Deprecation warnings to old style RecordDescriptor initialization methods
 * Added type hints and doc strings to RecordDescriptor class
 * RecordDescriptor.identifier is now a property to reduce hash calculation

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import collections
 import functools
 import gzip
@@ -570,7 +572,7 @@ class RecordDescriptor:
         """
         return self.init_from_dict(record._asdict(), raise_unknown=raise_unknown)
 
-    def extend(self, fields: Sequence[Tuple[str, str]]) -> "RecordDescriptor":
+    def extend(self, fields: Sequence[Tuple[str, str]]) -> RecordDescriptor:
         """Returns a new RecordDescriptor with the extended fields
 
         Returns:
@@ -614,7 +616,7 @@ class RecordDescriptor:
     def __hash__(self) -> int:
         return hash((self.name, self.get_field_tuples()))
 
-    def __eq__(self, other: "RecordDescriptor") -> bool:
+    def __eq__(self, other: RecordDescriptor) -> bool:
         if isinstance(other, RecordDescriptor):
             return self.name == other.name and self.get_field_tuples() == other.get_field_tuples()
         return NotImplemented
@@ -649,7 +651,7 @@ class RecordDescriptor:
         return (self.name, self._field_tuples)
 
     @staticmethod
-    def _unpack(name, fields: Tuple[Tuple[str, str]]) -> "RecordDescriptor":
+    def _unpack(name, fields: Tuple[Tuple[str, str]]) -> RecordDescriptor:
         return RecordDescriptor(name, fields)
 
 
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -10,7 +10,10 @@ import os
 import re
 import struct
 import sys
-from typing import Iterator, List, Optional, Tuple
+import warnings
+from datetime import datetime
+from itertools import zip_longest
+from typing import Any, Dict, Iterator, List, Mapping, Optional, Sequence, Tuple
 from urllib.parse import parse_qsl, urlparse
 
 from .exceptions import RecordDescriptorError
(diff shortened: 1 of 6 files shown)


flow/record/base.py
```diff
@@ -321,7 +318,7 @@ class RecordField:
     typename = None
     type = None
 
-    def __init__(self, name, typename):
+    def __init__(self, name: str, typename: str) -> None:
```
Schamper, line 321:
```suggestion
    def __init__(self, name: str, typename: str):
```
Return type of `__init__` can be skipped.

flow/record/base.py
```diff
@@ -338,243 +335,308 @@ class RecordFieldSet(list):
...
-        all_fields = self.get_all_fields()
         self.name = name
+        self._field_tuples = tuple([(to_native_str(k), to_str(v)) for k, v in fields])
+        self.recordType = _generate_record_class(name, self._field_tuples)
+        self.recordType._desc = self
 
-        if not RE_VALID_RECORD_TYPE_NAME.match(name):
-            raise RecordDescriptorError("Invalid record type name")
+    @staticmethod
+    @functools.lru_cache()
```
Schamper, line 470:
```suggestion
    @functools.cache()
```
I think you can just use `cache` here. Can't this be a constant, by the way?

flow/record/base.py
```diff
@@ -338,243 +335,308 @@ class RecordFieldSet(list):
...
+        }
 
         Returns:
-            OrderedDict
-
+            Mapping of all Record fields
         """
-        required_fields = OrderedDict([(k, RecordField(k, v)) for k, v in RESERVED_FIELDS.items()])
-        return required_fields
+        if self._all_fields is None:
```
Schamper, line 521: Is there a noticeable different between doing it like this and decorating this function with `@functools.cache`?

flow/record/base.py
```diff
@@ -338,243 +335,308 @@ class RecordFieldSet(list):
...
 
         Returns:
-            Record
-
+            Record with data from `record`
         """
         return self.init_from_dict(record._asdict(), raise_unknown=raise_unknown)
 
-    def extend(self, fields):
+    def extend(self, fields: Sequence[Tuple[str, str]]) -> "RecordDescriptor":
```
Schamper, line 573: You could `from __future__ import annotations` to allow using `RecordDescriptor` here.

Schamper (changes requested):
I was also thinking you might be able to cache the `fieldtype` function.

flow/record/base.py
yunzheng: flow.record still targets python 3.7 (by my request), so it still needs to be `lru_cache()`. The functionality is the same.

In my earlier refactor I turned this into the constant `REQUIRED_FIELDS`, but then I had to move things around, because it uses `RecordField()`. So the constant would disappear somewhere between methods, instead of below the currently defined constants such as `RESERVED_FIELDS`.

Also, there is an unit test that tests if you correctly use the required fields (_version always has to come last). It made testing that part a bit harder.

Let me know what you think.

flow/record/base.py
yunzheng: This is a class instance method, so it's not recommended to use `@functools.cache`. 

I know you also know this :) but for reference: https://rednafi.github.io/reflections/dont-wrap-instance-methods-with-functoolslru_cache-decorator-in-python.html

flow/record/base.py
yunzheng: I'll check if this is compatible with Python 3.7.

yunzheng: Fixed in 4609670

flow/record/base.py
Schamper: Right, forgot that `@cache` is 3.9.

Hm yeah that's annoying. I don't think this is a particularly hot function to perhaps better to just keep it this way.

yunzheng:
btw, I'm not sure if any `dissect` project use the old style RecordDescriptor initialization code.

Schamper:
> btw, I'm not sure if any `dissect` project use the old style RecordDescriptor initialization code.

Don't think so.

Schamper (approved):

PR #59, opened 2023-02-28
Support file-like inputs for RecordReader

MaxGroot:
- Peek into the file to find the right adapter
- Add tests for avro
- Add avro to testenv
- Add avro to extras in setup.py


flow/record/base.py
```diff
@@ -723,66 +789,73 @@ def open_path(path, mode, clobber=True):
...
-    adapter_scheme = ext_to_adapter.get(ext, "stream")
-    if "://" not in url:
-        url = f"{adapter_scheme}://{url}"
+        # First, we open the stream. If the stream is compressed, open_stream will wrap it for us into a decompressor.
+        cls_stream = Peekable(open_stream(url, "rb"))
 
-    p = urlparse(url, scheme=adapter_scheme)
-    adapter, _, sub_adapter = p.scheme.partition("+")
+        # We will again peek into the bytes to find an adapter, so we wrap the cls_stream into a Peekable object.
+        adapter = find_adapter_for_stream(cls_stream)
```
yunzheng, line 814: Would it be possible to move the `Peekable()` wrapper into `find_adapter_for_stream()`, like what you do in `open_stream()`.

flow/record/base.py
```diff
@@ -723,66 +789,73 @@ def open_path(path, mode, clobber=True):
...
 
     log.debug("Creating {!r} for {!r} with args {!r}".format(cls, url, arg_dict))
+
+    if cls_stream:
+        return cls(cls_stream, **arg_dict)
     return cls(cls_url, **arg_dict)
 
 
-def RecordReader(url=None, selector=None, **kwargs):
+def RecordReader(url=None, selector=None, **kwargs) -> AbstractReader:
```
yunzheng, line 854: Now `RecordReader()` can be called with both a url or a file-like object. If we keep it this way it would be better to also add type hints in the arguments to denote this.

The same applies for `RecordAdapter`. I'm not so sure about the API design, maybe a new wrapper function would be better, eg: `RecordFileReader(fp)` ? @Schamper what do you think?

flow/record/base.py
```diff
@@ -661,7 +679,55 @@ def DynamicDescriptor(name, fields):
     return RecordDescriptor(name, [("dynamic", field) for field in fields])
 
 
-def open_path(path, mode, clobber=True):
+def open_stream(fp, mode) -> IO | Peekable:
```
yunzheng, line 682: Maybe you can subclass `Peekable` from an IO class, then this type hint can be more generic / just IO..

flow/record/base.py
```diff
@@ -661,7 +679,55 @@ def DynamicDescriptor(name, fields):
...
+    elif HAS_LZ4 and peek_data[:4] == LZ4_MAGIC:
+        fp = lz4.open(fp, mode=mode)
+    elif HAS_ZSTD and peek_data[:4] == ZSTD_MAGIC:
+        dctx = zstd.ZstdDecompressor()
+        fp = dctx.stream_reader(fp)
+
+    return fp
+
+
+def find_adapter_for_stream(fp: Peekable) -> str:
```
yunzheng, line 704: maybe make this just fp: IO, and then check if you have `peek` and wrap accordingly. Like you do in `open_stream()`.

flow/record/base.py
```diff
@@ -723,66 +789,73 @@ def open_path(path, mode, clobber=True):
...
-                dctx = zstd.ZstdDecompressor()
-                fp = dctx.stream_reader(fp)
+            fp = open_stream(fp, mode)
     return fp
 
 
-def RecordAdapter(url, out, selector=None, clobber=True, **kwargs):
-    url = str(url or "")
-
+def RecordAdapter(url, out: bool, selector=None, clobber=True, **kwargs) -> AbstractWriter | AbstractReader:
```
yunzheng, line 796: The `|` type hint notation is only available in newer versions of Python. This should be `Union`.

yunzheng (changes requested):
I added some feedback

flow/record/base.py
Schamper: `BinaryIO` then, I guess?

Schamper (changes requested):

flow/record/adapter/avro.py
```diff
@@ -156,7 +157,7 @@ def descriptor_to_schema(desc):
     return schema
 
 
-def schema_to_descriptor(schema):
+def schema_to_descriptor(schema: dict) -> record.RecordDescriptor:
```
Schamper, line 160:
```suggestion
def schema_to_descriptor(schema: Dict) -> record.RecordDescriptor:
```
Otherwise I believe we break the 3.7 Python requirement of `flow.record`.

flow/record/adapter/avro.py
```diff
@@ -178,7 +179,7 @@ def schema_to_descriptor(schema):
     return record.RecordDescriptor(name, fields)
 
 
-def avro_type_to_flow_type(ftype):
+def avro_type_to_flow_type(ftype: list) -> str:
```
Schamper, line 182:
```suggestion
def avro_type_to_flow_type(ftype: List) -> str:
```

flow/record/adapter/stream.py
```diff
@@ -43,13 +45,13 @@ class StreamReader(AbstractReader):
     stream = None
 
     def __init__(self, path, selector=None, **kwargs):
```
Schamper, line 47: Can you add type hints to the arguments?

flow/record/adapter/stream.py
```diff
@@ -19,16 +21,16 @@ def __init__(self, path, clobber=True, **kwargs):
         self.fp = record.open_path(path, "wb", clobber=clobber)
         self.stream = record.RecordOutput(self.fp)
 
-    def write(self, r):
+    def write(self, r) -> None:
```
Schamper, line 24: Can you add a type hint to `r`?

flow/record/base.py
```diff
@@ -14,10 +14,25 @@
...
+    Sequence,
+    Tuple,
+    Union,
+)
 from urllib.parse import parse_qsl, urlparse
 
-from .exceptions import RecordDescriptorError
+from flow.record.adapter import AbstractReader, AbstractWriter
+
+from .exceptions import RecordAdapterNotFound, RecordDescriptorError
```
Schamper, line 35: Maybe make the import from `exceptions` also absolute.

flow/record/base.py
```diff
@@ -82,27 +108,31 @@ def _unpack(__cls, {args}):
 """
 
 
-class Peekable:
+class Peekable(BinaryIO):
     """Wrapper class for adding .peek() to a file object."""
 
     def __init__(self, fd):
```
Schamper, line 114:
```suggestion
    def __init__(self, fd: BinaryIO):
```

flow/record/base.py
```diff
@@ -82,27 +108,31 @@ def _unpack(__cls, {args}):
...
         self.fd = fd
         self.buffer = None
+        self.peeked_data = None
 
     def peek(self, size):
```
Schamper, line 119:
```suggestion
    def peek(self, size: int) -> bytes:
```

flow/record/base.py
```diff
@@ -82,27 +108,31 @@ def _unpack(__cls, {args}):
...
         if self.buffer is not None:
             raise BufferError("Only 1 peek allowed")
         data = self.fd.read(size)
         self.buffer = io.BytesIO(data)
+        self.peeked_data = data
```
Schamper, line 124: What is this needed for btw?

flow/record/base.py
```diff
@@ -82,27 +108,31 @@ def _unpack(__cls, {args}):
...
         return data
 
     def read(self, size=None):
```
Schamper, line 127:
```suggestion
    def read(self, size: Optional[int] = None) -> bytes:
```

flow/record/base.py
```diff
@@ -661,7 +691,56 @@ def DynamicDescriptor(name, fields):
...
+        fp = lz4.open(fp, mode=mode)
+    elif HAS_ZSTD and peek_data[:4] == ZSTD_MAGIC:
+        dctx = zstd.ZstdDecompressor()
+        fp = dctx.stream_reader(fp)
+
+    return fp
+
+
+def find_adapter_for_stream(fp: BinaryIO) -> tuple[BinaryIO, str]:
+    new_stream = Peekable(fp)
```
Schamper, line 717:
```suggestion
    if not hasattr(fp, "peek"):
        fp = Peekable(fp)
```
?

flow/record/base.py
```diff
@@ -661,7 +691,56 @@ def DynamicDescriptor(name, fields):
...
+    peek_data = new_stream.peek(RECORDSTREAM_MAGIC_DEPTH)
+    if new_stream.peeked_data is not None:
+        peek_data = new_stream.peeked_data
```
Schamper, line 720: Why not just call `.peek` again?

flow/record/base.py
```diff
@@ -661,7 +691,56 @@ def DynamicDescriptor(name, fields):
...
+    else:
+        peek_data = new_stream.peek(RECORDSTREAM_MAGIC_DEPTH)
+    if peek_data[:3] == AVRO_MAGIC and HAS_AVRO:
+        return new_stream, "avro"
+    elif RECORDSTREAM_MAGIC in peek_data:
+        return new_stream, "stream"
+    return None, None
+
+
+def open_file(path: Any, mode: str, clobber: bool = True) -> BinaryIO:
```
Schamper, line 730:
```suggestion
def open_file(path: Union[Path, str, BinaryIO], mode: str, clobber: bool = True) -> IO:
```
The return value should be `IO` I think because with `mode` I think you can open a text file?

flow/record/base.py
```diff
@@ -661,7 +691,56 @@ def DynamicDescriptor(name, fields):
...
+    if isinstance(path, Path):
+        path = str(path)
+    if isinstance(path, str):
+        return open_path(path, mode, clobber)
+    elif isinstance(path, Peekable):
```
Schamper, line 735: Would this not prevent any decompression from being opened?

flow/record/base.py
```diff
@@ -723,67 +802,89 @@ def open_path(path, mode, clobber=True):
...
-                fp = dctx.stream_reader(fp)
+            fp = open_stream(fp, mode)
     return fp
 
 
-def RecordAdapter(url, out, selector=None, clobber=True, **kwargs):
-    url = str(url or "")
-
+def RecordAdapter(
+    url: str = None, out: bool = False, selector=None, clobber: bool = True, fileobj: BinaryIO = None, **kwargs
```
Schamper, line 810:
```suggestion
    url: str = None, out: bool = False, selector: Optional[str] = None, clobber: bool = True, fileobj: Optional[BinaryIO] = None, **kwargs
```
I believe `selector` is a `str`.

flow/record/base.py
```diff
@@ -723,67 +802,89 @@ def open_path(path, mode, clobber=True):
...
-
-    p = urlparse(url, scheme=adapter_scheme)
-    adapter, _, sub_adapter = p.scheme.partition("+")
-
+    cls_stream = None
+    cls_url = None
+    adapter = None
+
+    # When a url is given, we interpret it to determine what kind of adapter we need. This piece of logic is always
+    # necessary for the RecordWriter (as it does not currently support filelike objects), and only needed for
```
Schamper, line 824:
```suggestion
    # necessary for the RecordWriter (as it does not currently support file-like objects), and only needed for
```

flow/record/base.py
```diff
@@ -723,67 +802,89 @@ def open_path(path, mode, clobber=True):
...
+        arg_dict.update(kwargs)
+
+        cls_url = p.netloc + p.path
+        if sub_adapter:
+            cls_url = sub_adapter + "://" + cls_url
+
+    if fileobj is not None:
+        # This record adapter has received a file-like object.
+        # We just need to find the right adapter
+        # by peeking into the first few bytes.
```
Schamper, line 847:
```suggestion
        # We just need to find the right adapter by peeking into the first few bytes.
```

flow/record/base.py
```diff
@@ -723,67 +802,89 @@ def open_path(path, mode, clobber=True):
...
     log.debug("Creating {!r} for {!r} with args {!r}".format(cls, url, arg_dict))
+
+    if fileobj is not None:
+        return cls(cls_stream, **arg_dict)
     return cls(cls_url, **arg_dict)
 
 
-def RecordReader(url=None, selector=None, **kwargs):
-    return RecordAdapter(url, False, selector=selector, **kwargs)
+def RecordReader(url: str = None, selector: str = None, fileobj: BinaryIO = None, **kwargs) -> AbstractReader:
```
Schamper, line 882:
```suggestion
def RecordReader(url: Optional[str] = None, selector: Optional[str] = None, fileobj: Optional[BinaryIO] = None, **kwargs) -> AbstractReader:
```

flow/record/base.py
```diff
@@ -723,67 +802,89 @@ def open_path(path, mode, clobber=True):
...
+    return RecordAdapter(url=url, out=False, selector=selector, fileobj=fileobj, **kwargs)
 
 
-def RecordWriter(url=None, clobber=True, **kwargs):
-    return RecordAdapter(url, True, clobber=clobber, **kwargs)
+def RecordWriter(url: str = None, clobber: bool = True, **kwargs) -> AbstractWriter:
```
Schamper, line 886:
```suggestion
def RecordWriter(url: Optional[str] = None, clobber: bool = True, **kwargs) -> AbstractWriter:
```

flow/record/base.py
```diff
@@ -723,67 +802,89 @@ def open_path(path, mode, clobber=True):
...
+    return RecordAdapter(url=url, out=True, clobber=clobber, fileobj=None, **kwargs)
```
Schamper, line 887:
```suggestion
    return RecordAdapter(url=url, out=True, clobber=clobber, **kwargs)
```

Schamper (changes requested):

flow/record/base.py
MaxGroot: You were right, this wasn't needed as we could use the buffer property.

flow/record/base.py
MaxGroot: As calling `.peek` on the same `Peekable` instance isn't allowed, I initially opted for the `peeked_data` variable. However, this is unnecessarily complex, so I opted to just always wrap the incoming `fp` within a new `Peekable` instance for the `find_adapter_for_stream` function.

flow/record/base.py
MaxGroot: Could you elaborate in what situation this would prevent decompression from being opened? As far as I understand, the existing tests already cover compression and decompression for paths, and I added tests for decompression of file-like objects, and I didn't encounter a scenario where decompression failed.

flow/record/base.py
```diff
@@ -83,27 +108,29 @@ def _unpack(__cls, {args}):
...
         return data
 
-    def read(self, size=None):
-        data = b""
+    def read(self, size: Optional[int] = None) -> bytes:
         if self.buffer is None:
-            data = self.fd.read(size)
+            return self.fd.read(size)
         else:
+            data = b""
```
yunzheng, line 129: `data = b""` looks to redundant and can be removed.

flow/record/base.py
```diff
@@ -662,7 +689,56 @@ def DynamicDescriptor(name, fields):
...
+
+
+def find_adapter_for_stream(fp: BinaryIO) -> tuple[Peekable, str]:
+    # We need to peek into the stream to be able to determine which adapter is needed. The fp given to this function
+    # might already be an instance of the 'Peekable' class, but might also be a different file pointer, for example
+    # a transparent decompressor. As calling peek() twice on the same peekable is not allowed, we wrap the fp into
+    # a Peekable again, so that we are able to determine the correct adapter.
+    fp = Peekable(fp)
+    peek_data = fp.peek(RECORDSTREAM_MAGIC_DEPTH)
+    if peek_data[:3] == AVRO_MAGIC and HAS_AVRO:
```
yunzheng, line 721: Move HAS_AVRO as first check

```suggestion
    if HAS_AVRO and peek_data[:3] == AVRO_MAGIC:
```

flow/record/base.py
```diff
@@ -724,67 +800,109 @@ def open_path(path, mode, clobber=True):
...
+        # If a user did not provide a url, we have to peek into the stream to be able to determine the right adapter
+        # based on magic bytes encountered in the first few bytes of the stream.
+        if adapter is None:
+            cls_stream, adapter = find_adapter_for_stream(cls_stream)
+            if adapter is None:
+                if hasattr(cls_stream, "buffer") and cls_stream.buffer.getvalue().startswith(b"<"):
+                    raise RecordAdapterNotFound(
+                        (
+                            "Could not find a reader for this input. Are you perhaps entering record text,"
+                            " rather than a recordstream? This can be fixed by using 'rdump -w -' to write a"
```
yunzheng, line 872: Took me a while to figure out where buffer.getvalue() comes from, but this is because of the Peekable buffer and only available when a `.peek()` was done.

Added a small preview of the peeked buffer to the error message:

```suggestion
                            f"Could not find a reader for input {cls_stream.buffer.getvalue()!r}. Are you perhaps entering record text,"
```

yunzheng (changes requested):
Some small suggestions and feedback

tests/test_rdump.py
```diff
@@ -461,6 +461,46 @@ def test_rdump_headerless_csv(tmp_path, capsysbinary):
...
+        writer.write(TestRecord(count=i, foo="bar"))
+    writer.close()
+
+    # Compress
+    compress_cmd = ["gzip", "/k", str(path)]
+
+    res = subprocess.Popen(compress_cmd, stdout=subprocess.PIPE)
+    res.communicate()
+
+    compressed_path = str(path) + ".gz"
```
yunzheng, line 489: `gzip /k` should be `gzip -k/--keep`:

```suggestion
    # Gzip compress records file
    compress_cmd = ["gzip", "--keep", str(path)]
    subprocess.check_output(compress_cmd)
    compressed_path = str(path) + ".gz"
```

yunzheng (changes requested):
Small unit test issue, then it looks good to go

flow/record/base.py
```diff
@@ -83,27 +108,28 @@ def _unpack(__cls, {args}):
 """
 
 
-class Peekable:
+class Peekable(BinaryIO):
```
Schamper, line 111: I wonder if [io.BufferedReader](https://docs.python.org/3/library/io.html#io.BufferedReader) could work instead of this custom class? Also, you'd probably want to subclass from `io.RawIOBase` instead of `BinaryIO`, which is a type hint.

flow/record/base.py
```diff
@@ -662,7 +688,56 @@ def DynamicDescriptor(name, fields):
...
+    elif HAS_LZ4 and peek_data[:4] == LZ4_MAGIC:
+        fp = lz4.open(fp, mode=mode)
+    elif HAS_ZSTD and peek_data[:4] == ZSTD_MAGIC:
+        dctx = zstd.ZstdDecompressor()
+        fp = dctx.stream_reader(fp)
+
+    return fp
+
+
+def find_adapter_for_stream(fp: BinaryIO) -> tuple[Peekable, str]:
```
Schamper, line 713:
```suggestion
def find_adapter_for_stream(fp: BinaryIO) -> tuple[BinaryIO, Optional[str]]:
```

flow/record/base.py
```diff
@@ -662,7 +688,56 @@ def DynamicDescriptor(name, fields):
...
+    fp = Peekable(fp)
+    peek_data = fp.peek(RECORDSTREAM_MAGIC_DEPTH)
+    if HAS_AVRO and peek_data[:3] == AVRO_MAGIC:
+        return fp, "avro"
+    elif RECORDSTREAM_MAGIC in peek_data:
+        return fp, "stream"
+    return fp, None
+
+
+def open_file(path: Any, mode: str, clobber: bool = True) -> IO:
```
Schamper, line 727:
```suggestion
def open_file(path: Union[str, Path, BinaryIO], mode: str, clobber: bool = True) -> IO:
```
?

flow/record/base.py
Schamper: I haven't followed all code paths, but exiting early here in case of a `Peekable` and not calling `open_stream`, I figured that if you somehow where to have a `Peekable` of a (still) compressed stream, we wouldn't do compression stream detection on it. This is probably not a realistic code path though.

flow/record/base.py
```diff
@@ -662,7 +688,56 @@ def DynamicDescriptor(name, fields):
...
+        return path
+    elif isinstance(path, io.IOBase):
+        return open_stream(path, "rb")
+    else:
+        raise ValueError(f"Unsupported path type {path}")
+
+
+def open_path(path: str, mode: str, clobber: bool = True) -> IO:
     """
     Open `path` using `mode` and returns a file object.
```
Schamper, line 742:
```suggestion
    Open ``path`` using ``mode`` and returns a file object.
```
Some quick Sphinx fixes while we're here. Also there's an additional newline at the end of this docstring. And you could remove the type hints in the docstring now that they're properly added. Though this could be changed some other time.

flow/record/base.py
```diff
@@ -724,67 +799,109 @@ def open_path(path, mode, clobber=True):
...
-                fp = dctx.stream_reader(fp)
+            fp = open_stream(fp, mode)
     return fp
 
 
-def RecordAdapter(url, out, selector=None, clobber=True, **kwargs):
-    url = str(url or "")
-
+def RecordAdapter(
+    url: str = None,
```
Schamper, line 807:
```suggestion
    url: Optional[str] = None,
```

flow/record/base.py
```diff
@@ -724,67 +799,109 @@ def open_path(path, mode, clobber=True):
...
+        # based on magic bytes encountered in the first few bytes of the stream.
+        if adapter is None:
+            cls_stream, adapter = find_adapter_for_stream(cls_stream)
+            if adapter is None:
+                if hasattr(cls_stream, "buffer") and cls_stream.buffer.getvalue().startswith(b"<"):
+                    raise RecordAdapterNotFound(
+                        (
+                            f"Could not find a reader for input {cls_stream.buffer.getvalue()!r}. Are you perhaps "
+                            "entering record text, rather than a recordstream? This can be fixed by using 'rdump -w -' "
+                            "to write a recordstream to stdout."
```
Schamper, line 872:
```suggestion
                            "entering record text, rather than a record stream? This can be fixed by using 'rdump -w -' "
                            "to write a record stream to stdout."
```
Seems weird to put it as one word.

Schamper (changes requested):

flow/record/base.py
MaxGroot: Could you elaborate on the change from `Peekable` to `BinaryIO`? In its current form, the `find_adapter_for_stream` function always wraps the given `fp` in a `Peekable` object before returning it.

Schamper: Does it matter for the return value though? Is functionality of `Peekable` ever required afterwards, or should it just behave as a `BinaryIO`?

MaxGroot: `io.BufferedReader` seems to be just fine so I'll make a commit that replaces the `Peekable` class with usage of `io.BufferedReader`. Thus you can ignore my previous comment.

flow/record/adapter/avro.py
```diff
@@ -117,13 +118,13 @@ def __iter__(self):
...
 
-    def close(self):
+    def close(self) -> None:
         if self.fp:
             self.fp.close()
         self.fp = None
 
 
-def descriptor_to_schema(desc):
+def descriptor_to_schema(desc: record.RecordDescriptor) -> Dict[str, Any]:
```
Schamper, line 127:
```suggestion
def descriptor_to_schema(desc: record.RecordDescriptor) -> dict[str, Any]:
```
This should be legal if you add a `from __future__ import annotations` at the top.

tests/test_avro.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import BytesIO as StringIO
```
Schamper, line 6: I don't think we want to add new Python2 compatibility code anymore.

Schamper (changes requested):

flow/record/base.py
```diff
@@ -14,10 +14,24 @@
...
+    IO,
+    Any,
+    BinaryIO,
+    Dict,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
```
Schamper, line 28: Can you replace all occurrences of `Dict`, `List` and `Tuple` with `dict`, `list` and `tuple`?

Schamper (changes requested):

flow/record/base.py
```diff
@@ -578,19 +566,19 @@ def extend(self, fields: Sequence[Tuple[str, str]]) -> RecordDescriptor:
...
 
-    def get_field_tuples(self) -> Tuple[Tuple[str, str]]:
+    def get_field_tuples(self) -> tuple[tuple[str, str]]:
         """Returns a tuple containing the (typename, name) tuples, eg:
 
         (('boolean', 'foo'), ('string', 'bar'))
 
         Returns:
-            Tuple of (typename, name) tuples
+            tuple of (typename, name) tuples
```
Schamper, line 575:
```suggestion
            Tuple of (typename, name) tuples
```
Bit too eager find & replace.

flow/record/base.py
```diff
@@ -923,15 +911,15 @@ def fieldtype(clspath: str) -> FieldType:
...
+    descriptors: tuple[RecordDescriptor], replace: bool = False, name: Optional[str] = None
 ) -> RecordDescriptor:
     """Create a newly merged RecordDescriptor from a list of RecordDescriptors.
     This function uses a cache to avoid creating the same descriptor multiple times.
 
     Duplicate fields are ignored in ``descriptors`` unless ``replace=True``.
 
     Args:
-        descriptors: Tuple of RecordDescriptors to merge.
+        descriptors: tuple of RecordDescriptors to merge.
```
Schamper, line 922:
```suggestion
        descriptors: Tuple of RecordDescriptors to merge.
```

flow/record/base.py
```diff
@@ -950,15 +938,15 @@ def merge_record_descriptors(
...
+    record: Record, other_records: list[Record], replace: bool = False, name: Optional[str] = None
 ) -> Record:
     """Extend ``record`` with fields and values from ``other_records``.
 
     Duplicate fields are ignored in ``other_records`` unless ``replace=True``.
 
     Args:
         record: Initial Record to extend.
-        other_records: List of Records to use for extending/replacing.
+        other_records: list of Records to use for extending/replacing.
```
Schamper, line 949:
```suggestion
        other_records: List of Records to use for extending/replacing.
```

Schamper (changes requested):

Schamper (approved):

yunzheng (approved):
LGTM

PR #61, opened 2023-03-16
Add UTC timezone when converting timestamps

JSCU-CNI:
This PR changes `datetime` fieldtype behaviour when dealing with input integers and floats.

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -244,7 +244,9 @@ class datetime(_dt, FieldType):
                 else:
                     return cls.fromisoformat(arg)
             elif isinstance(arg, (int, float_type)):
-                return cls.utcfromtimestamp(arg)
+                # When we receive integers and floats we expect
+                # them to be epoch timestamps which are UTC by definition.
+                return cls.fromtimestamp(arg, tz=timezone.utc)
             elif isinstance(arg, (_dt,)):
                 return _dt.__new__(
                     cls,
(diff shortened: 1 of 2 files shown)


yunzheng:
flow.record internally doesn't use timezone aware datetime objects, but it does support it. Internally when the datetime has no timezone info (naive datetime) it is always considered in UTC. So the same applies when loading from float or integers.

However, timezone suport in flow.record can be better. For example it will currently break if you compare between naive and aware flow.record datetime fields.  Was this the reason for this change/PR?

Currently it only supports `__eq__` so it should be expanded to also other comparison methods.
We have also been thinking to make the datetime fields timezone aware by default, but we noticed there is a significant slowdown when working with timezone aware datetime fields.

JSCU-CNI:
Sorry for our late reply! :blush:

The reason for submitting this PR was indeed that we came across an instance in which some flow.record objects were timezone aware and some were naive. Since naive and aware datetime objects cannot be compared (which is a good thing in our opinion), it is impossible to relate events to each other based on datetime values.

We have put in quite some effort to make all dissect.target plugins timezone-aware (e.g. PRs [#116](https://github.com/fox-it/dissect.target/pull/116) and [#150](https://github.com/fox-it/dissect.target/pull/150)). Moreover, all other methods for initializing a flow.record datetime object already do support timezones. Therefore, this PR makes the initialization of datetime objects more uniform.

yunzheng:
Thanks, also sorry for my late reply :)

Maybe it's time for `flow.record` to move completely to native timezone aware `datetime` as a default. I will discuss this internally, as I have tried to work on this before but came across performance issues.

If we decide that we want to do this, this PR can be merged before we fix other places that also need to be made compatible.

yunzheng:
Good news, we did some benchmarks and testing and concluded that we can support timezone aware datetimes without too much performance hits.

I'm closing this PR in favour for #78, please check if this fixes the issues or if you have further comments.

PR #62, opened 2023-03-16
Always write Avro header

Schamper:
Initialize an empty Avro reader if we flush with no writer initialized yet. This ensures that the Avro header is always written so that the resulting file is a valid, but empty, Avro file.


yunzheng (changes requested):
Can you add a small unit test?

PR #65, opened 2023-04-20
Allow datetime to process strings with nanoseconds

cecinestpasunepipe:
(DIS-1964)

### DIFF
--- a/tests/test_fieldtypes.py
+++ b/tests/test_fieldtypes.py
@@ -9,13 +9,9 @@ import pytest
 
 import flow.record.fieldtypes
 from flow.record import RecordDescriptor, RecordReader, RecordWriter
-from flow.record.fieldtypes import (
-    PATH_POSIX,
-    PATH_WINDOWS,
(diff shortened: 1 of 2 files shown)


tests/test_fieldtypes.py
```diff
@@ -9,13 +9,9 @@
...
-from flow.record.fieldtypes import (
-    PATH_POSIX,
-    PATH_WINDOWS,
-    fieldtype_for_value,
-    net,
-    uri,
-)
+from flow.record.fieldtypes import PATH_POSIX, PATH_WINDOWS
+from flow.record.fieldtypes import datetime as dt
+from flow.record.fieldtypes import fieldtype_for_value, net, uri
```
pyrco, line 14: Why this change? isort or black don't like the tuple formatting when `as` is used?

cecinestpasunepipe: Yes, that was isort I'm afraid...

pyrco (approved):

PR #66, opened 2023-04-24
Improve the path field type

pyrco:
The path class will now allow being instantiated using multiple path parts, making it behave more like its parent pathlib.PurePath class.

The inference on whether to return a windows_path or posix_path class instance is
now done based on the first path part it encounters that is an instance of pathlib.PurePath.

For path parts that are a custom subclass of pathlib.PurePath (so not an instance of PureWindowsPath or PurePosixPath), the presence of a '\' as either separator or alternative separator will result in returning a windows_path class instance.

(DIS-1977)


flow/record/fieldtypes/__init__.py
```diff
@@ -542,23 +542,64 @@ def _unpack(cls, data):
         return data
 
 
+def _is_posixlike_path(path: Any):
+    return isinstance(path, pathlib.PurePath) and path._flavour.sep != "\\" and path._flavour.altsep != "\\"
```
cecinestpasunepipe, line 546:
```suggestion
    return isinstance(path, pathlib.PurePath) and "\\" not in (path._flavour.sep, path._flavour.altsep)
```

flow/record/fieldtypes/__init__.py
```diff
@@ -542,23 +542,64 @@ def _unpack(cls, data):
...
+
+
+def _is_windowslike_path(path: Any):
+    return isinstance(path, pathlib.PurePath) and (path._flavour.sep == "\\" or path._flavour.altsep == "\\")
```
cecinestpasunepipe, line 550:
```suggestion
    return isinstance(path, pathlib.PurePath) and "\\" in (path._flavour.sep, path._flavour.altsep)
```

flow/record/fieldtypes/__init__.py
```diff
@@ -542,23 +542,64 @@ def _unpack(cls, data):
...
-                    args = [path_.as_posix()]
-                elif isinstance(path_, pathlib.PurePosixPath):
+                    args = (str(arg) if _is_windowslike_path(arg) else arg for arg in args)
+                elif _is_posixlike_path(path_part):
+                    # This handles any custom PurePath based implementations that don't have a
+                    # windows like path separator (\).
                     cls = posix_path
-                    args = [path_.as_posix()]
+                else:
+                    continue
```
cecinestpasunepipe, line 596: What happens here? Is this a realistic case, i.e. we don't know? Why make a loop anyway instead of just picking the first element from args?

cecinestpasunepipe (changes requested):

flow/record/fieldtypes/__init__.py
Schamper: Note that this might be slower.

flow/record/fieldtypes/__init__.py
Schamper: Note that this might be slower

pyrco: It will probably be a fraction slower due to the construction of a tuple. Will this be an issue? the `path` field type is already relatively heavy.

flow/record/fieldtypes/__init__.py
pyrco: The type of path is determined by the first path part from which such a thing can be inferred. In the case of this implementation that means being a subclass of `pathlib.PurePath`.

As path parts are also allowed to be plain `str` types (or `PathLike`, which amounts to being str typed), we have to iterate over the parts until we find the first part that has a type we are looking for.  The only Python types that can make explicit how paths are formatted are `PurePath` subclasses. So we have to iterate and discard all path parts that are not a `PurePath` subclass.

The native `PurePath` doesn't do any inference at all, as it will always return a path interpreted according to how the local system defines paths. If the user doesn't want that behavior, they always have to explicitly use the correct `PurePath` subclass.

For the flow.record `path` class we do some form of auto detection for convenience, as flow.record is often used in systems that deal with paths which are not formatted according to the local system specifications.

flow/record/fieldtypes/__init__.py
cecinestpasunepipe: image

Does not seem to be so, at least in my simplified test case. But even if it is/was then we are coding against implementation details which might change from time to time, depending on internal Python code. Anyway, if you do not want the change it's okay, I just thought it would be a bit cleaner not to repeat the string inside that statement. But the benefits are quite limited either way.

flow/record/fieldtypes/__init__.py
Schamper: See my reply to the other thread.

cecinestpasunepipe (dismissed):

flow/record/fieldtypes/__init__.py
```diff
@@ -547,23 +547,63 @@ def _unpack(cls, data):
...
+                    # str(PureWindowsPath(PureWindowsPath(r"\C:\WINDOWS/Temp"))),
+                    #
+                    # repr(PureWindowsPath(r"\C:\WINDOWS/Temp")) !=
+                    # repr(PureWindowsPath(PureWindowsPath(r"\C:\WINDOWS/Temp"))),
+                    #
+                    # This would be the case though when using PurePosixPath instead.
+                    #
+                    # This construction works around that by converting all path parts that are
+                    # pathlib.PureWindowsPath descendants to strings first.
+                    args = (str(arg) if isinstance(arg, pathlib.PureWindowsPath) else arg for arg in args)
```
yunzheng, line 587: After determining the first path type, does converting all args to `str` work instead of specifically checking if it's a different type?

pyrco: Yes that would work. Technically it would work when if only the very first part is a `PureWindowsPath` only that part is converted to string.

I guess the idea is to make things a bit faster? I'll check the speed differences between casting everything to string vs converting args to a list and only converting the first part.

pyrco: I ran some tests and as expected, the option to always convert all parts if the first _encountered_ `PurePath` subclass is a `WindowsPurePath` is more or less constant time.

If we only convert the very _first_ part if it is a `WindowsPurePath` there is a slight slow down (due to an `enumerate` and extra `if` statement. If however The first part is not a `WindowsPurePath` it is slightly faster.

```
all_to_string ; path parts: 2 ; time: 0.5000176659996214
all_to_string ; path parts: 5 ; time: 0.4989467599989439
all_to_string ; path parts: 10 ; time: 0.5002165209989471
all_to_string ; path parts: 3 ; time: 0.5717005929982406
all_to_string ; path parts: 6 ; time: 0.5740014379989589
all_to_string ; path parts: 11 ; time: 0.572631115999684

first_to_string ; path parts: 2 ; time: 0.6870770460009226
first_to_string ; path parts: 5 ; time: 0.6977602070001012
first_to_string ; path parts: 10 ; time: 0.7180876769998576
first_to_string ; path parts: 3 ; time: 0.43469748999996227
first_to_string ; path parts: 6 ; time: 0.42878352199841174
first_to_string ; path parts: 11 ; time: 0.43002194500149926
```

Here the 2, 5 and 10 path parts have a `WindowsPurePath` as first part and the 3, 6 and 11 a string.

yunzheng: I would prefer the `all_string` for consistency and predictability in this case.

yunzheng (approved):
LGTM

PR #67, opened 2023-04-28
Fixed Value Error in filedtypes.datetime.

kollee:
filedtypes.datetime throws a Value Error if args contains a datetime string without timezone but microseconds. The fix ensures that existing microseconds are removed from the string before passing it to fromisoformat().

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -247,7 +247,8 @@ class datetime(_dt, FieldType):
                             return cls.strptime(re.sub(RE_STRIP_NANOSECS, "\\1", arg), "%Y-%m-%dT%H:%M:%S.%f%z")
                     return cls.strptime(arg, "%Y-%m-%dT%H:%M:%S%z")
                 else:
-                    return cls.fromisoformat(arg)
+                    # Sometimes nanoseconds need to be stripped
+                    return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "", arg))
             elif isinstance(arg, (int, float_type)):
                 return cls.utcfromtimestamp(arg)
             elif isinstance(arg, (_dt,)):
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -251,7 +251,7 @@ class datetime(_dt, FieldType):
                         return cls.fromisoformat(arg)
                     except ValueError:
                         # Sometimes nanoseconds need to be stripped
-                        return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "", arg))
+                        return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "\\1", arg))
             elif isinstance(arg, (int, float_type)):
                 return cls.utcfromtimestamp(arg)
             elif isinstance(arg, (_dt,)):
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -247,8 +247,11 @@ class datetime(_dt, FieldType):
                             return cls.strptime(re.sub(RE_STRIP_NANOSECS, "\\1", arg), "%Y-%m-%dT%H:%M:%S.%f%z")
                     return cls.strptime(arg, "%Y-%m-%dT%H:%M:%S%z")
                 else:
-                    # Sometimes nanoseconds need to be stripped
-                    return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "", arg))
+                    try:
+                        return cls.fromisoformat(arg)
+                    except ValueError:
+                        # Sometimes nanoseconds need to be stripped
+                        return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "", arg))
             elif isinstance(arg, (int, float_type)):
                 return cls.utcfromtimestamp(arg)
             elif isinstance(arg, (_dt,)):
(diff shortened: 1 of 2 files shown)


flow/record/fieldtypes/__init__.py
```diff
@@ -247,7 +247,8 @@ def __new__(cls, *args, **kwargs):
                             return cls.strptime(re.sub(RE_STRIP_NANOSECS, "\\1", arg), "%Y-%m-%dT%H:%M:%S.%f%z")
                     return cls.strptime(arg, "%Y-%m-%dT%H:%M:%S%z")
                 else:
-                    return cls.fromisoformat(arg)
+                    # Sometimes nanoseconds need to be stripped
+                    return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "", arg))
```
cecinestpasunepipe, line 251: Most of the time nanosecs are not present, since this code is quite hot maybe consider using a try-catch like with the timezone version? Also, in future versions of Python this is very likely to be fixed (although I have to confess I have not checked the roadmap, it just seems logical) So this might be a good reason to only apply this fix in case of a ValueError?

cecinestpasunepipe (changes requested):

flow/record/fieldtypes/__init__.py
kollee: I changed the code accordingly. Any further comments?

cecinestpasunepipe: No, looks good, thanks. But can you fix the linting issue please?

tests/test_fieldtypes.py
```diff
@@ -806,5 +807,13 @@ def test_datetime_strip_nanoseconds():
     assert d1 == d2
 
 
+def test_datetime_strip_nanoseconds_without_timezone():
+    d1 = dt("2006-11-10T14:29:55.5851926")
+    d2 = dt("2006-11-10T14:29:55")
+    assert isinstance(d1, dt)
+    assert isinstance(d2, dt)
+    assert d1 == d2
```
yunzheng, line 815: I expect still to have microseconds. So `585192` as microseconds.

yunzheng (changes requested):
Hi @kollee, thanks for finding the issue and making a PR!

I think the behaviour that it should still try to parse microseconds, the best way is to compare it to Python 3.11's  `fromisoformat()` function:

On python 3.11 this will work:

```python
Python 3.11.2 (main, Feb 11 2023, 02:24:27) [GCC 10.2.1 20210110]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.11.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import datetime

In [2]: datetime.datetime.fromisoformat("2006-11-10T14:29:55.5851926")
Out[2]: datetime.datetime(2006, 11, 10, 14, 29, 55, 585192)

In [3]: _.microsecond
Out[3]: 585192
```

However on older Python versions this will not work but it would preferred it behaves the same and returns the same values.

tests/test_fieldtypes.py
```diff
@@ -806,5 +807,13 @@ def test_datetime_strip_nanoseconds():
     assert d1 == d2
 
 
+def test_datetime_strip_nanoseconds_without_timezone():
+    d1 = dt("2006-11-10T14:29:55.5851926")
```
yunzheng, line 811: I expect still to have microseconds. So `585192` as microseconds.

yunzheng:
Hi @kollee,

I made the fixes to retain microseconds, but I could not push to your fork as it was not writeable.
If you can apply the following patch and mark the comments as resolved, it should be good!

```patch
diff --git a/flow/record/fieldtypes/__init__.py b/flow/record/fieldtypes/__init__.py
index e7c46bf..70a2f21 100644
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -251,7 +251,7 @@ class datetime(_dt, FieldType):
                         return cls.fromisoformat(arg)
                     except ValueError:
                         # Sometimes nanoseconds need to be stripped
-                        return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "", arg))
+                        return cls.fromisoformat(re.sub(RE_STRIP_NANOSECS, "\\1", arg))
             elif isinstance(arg, (int, float_type)):
                 return cls.utcfromtimestamp(arg)
             elif isinstance(arg, (_dt,)):
diff --git a/tests/test_fieldtypes.py b/tests/test_fieldtypes.py
index cfeb310..b9fece4 100644
--- a/tests/test_fieldtypes.py
+++ b/tests/test_fieldtypes.py
@@ -429,7 +429,8 @@ def test_datetime():
         ("2019-09-26T07:58:30.996+0200", datetime.datetime(2019, 9, 26, 5, 58, 30, 996000)),
         ("2011-11-04T00:05:23+04:00", datetime.datetime(2011, 11, 3, 20, 5, 23)),
         ("2023-01-01T12:00:00+01:00", datetime.datetime(2023, 1, 1, 11, 0, 0, tzinfo=datetime.timezone.utc)),
-        ("2006-11-10T14:29:55.5851926", datetime.datetime(2006, 11, 10, 14, 29, 55)),
+        ("2006-11-10T14:29:55.5851926", datetime.datetime(2006, 11, 10, 14, 29, 55, 585192)),
+        ("2006-11-10T14:29:55.585192699999999", datetime.datetime(2006, 11, 10, 14, 29, 55, 585192)),
         (datetime.datetime(2023, 1, 1, tzinfo=datetime.timezone.utc), datetime.datetime(2023, 1, 1)),
         (0, datetime.datetime(1970, 1, 1, 0, 0)),
     ],
@@ -812,7 +813,10 @@ def test_datetime_strip_nanoseconds_without_timezone():
     d2 = dt("2006-11-10T14:29:55")
     assert isinstance(d1, dt)
     assert isinstance(d2, dt)
-    assert d1 == d2
+    assert d1 == datetime.datetime(2006, 11, 10, 14, 29, 55, 585192)
+    assert d1.microsecond == 585192
+    assert d2 == datetime.datetime(2006, 11, 10, 14, 29, 55)
+    assert d2.microsecond == 0


 if __name__ == "__main__":
```

kollee:
> Hi @kollee,
> 
> I made the fixes to retain microseconds, but I could not push to your fork as it was not writeable. If you can apply the following patch and mark the comments as resolved, it should be good!

Hi @yunzheng,

you are of course right. I added you suggestions. 

Thank you all for providing dissect to the community.

yunzheng (approved):

yunzheng:
@cecinestpasunepipe if you have no further review requests, we can merge this

PR #76, opened 2023-07-11
Add skip flag to rdump

MaxGroot:
Adds the ability to skip a number of records when reading or writing records using `rdump`.

The added test might be a bit overkill, open to feedback as to how it can be simplified.

### DIFF
--- a/flow/record/tools/rdump.py
+++ b/flow/record/tools/rdump.py
@@ -205,6 +205,7 @@ def main(argv=None):
     seen_desc = set()
     islice_stop = (args.count + args.skip) if args.count else None
     record_iterator = islice(record_stream(args.src, selector), args.skip, islice_stop)
+    count = 0
     with RecordWriter(uri) as record_writer:
         for count, rec in enumerate(record_iterator, start=1):
             if args.record_source is not None:
--- a/flow/record/tools/rdump.py
+++ b/flow/record/tools/rdump.py
@@ -95,6 +95,7 @@ def main(argv=None):
     output = parser.add_argument_group("output control")
     output.add_argument("-f", "--format", metavar="FORMAT", help="Format string")
     output.add_argument("-c", "--count", type=int, help="Exit after COUNT records")
+    output.add_argument("--skip", metavar="COUNT", default=None, type=int, help="Skip the first COUNT records")
     output.add_argument("-w", "--writer", metavar="OUTPUT", default=None, help="Write records to output")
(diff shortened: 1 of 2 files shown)


flow/record/tools/rdump.py
```diff
@@ -95,6 +95,7 @@ def main(argv=None):
     output = parser.add_argument_group("output control")
     output.add_argument("-f", "--format", metavar="FORMAT", help="Format string")
     output.add_argument("-c", "--count", type=int, help="Exit after COUNT records")
+    output.add_argument("--skip", metavar="COUNT", default=None, type=int, help="Skip the first COUNT records")
```
Schamper, line 98:
```suggestion
    output.add_argument("--skip", metavar="COUNT", default=0, type=int, help="Skip the first COUNT records")
```

Schamper (dismissed):

flow/record/tools/rdump.py
```diff
@@ -202,8 +203,11 @@ def main(argv=None):
     selector = make_selector(args.selector, not args.no_compile)
     seen_desc = set()
     count = 0
+    skip = args.skip if args.skip else 0
     with RecordWriter(uri) as record_writer:
         for count, rec in enumerate(record_stream(args.src, selector), start=1):
+            if count <= skip:
```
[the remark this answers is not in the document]
Schamper: I think this actually also skips the first record. Perhaps something that can be easily remedied by using `itertools.islice`?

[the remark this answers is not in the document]
MaxGroot: I think that the `start=1` parameter setting for enumerate causes it to not skip the first record. Having said that, I'll look into `itertools.islice`.

flow/record/tools/rdump.py
```diff
@@ -202,8 +203,11 @@ def main(argv=None):
...
+                continue
```
yunzheng, line 210: Because `skip` is already the count of records you want to skip, you could also opt to just decrease it as a counter:

```suggestion
            if skip > 0:
                skip -= 1
                continue
```

yunzheng: This however does break `--count`, as the count is derived from `enumerate()`. A solution would be to get rid of `enumerate()` and track `count` separately when outputting records.

Schamper: `itertools.islice` solves all of these problems i believe.

yunzheng (approved):
LGTM

PR #77, opened 2023-07-19
Add hash method to datetime fieldtype

Poeloe:
(DIS-2036)

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -284,7 +284,7 @@ class datetime(_dt, FieldType):
         return result
 
     def __hash__(self):
-        return super.__hash__(self)
+        return _dt.__hash__(self)
 
 
 class varint(varint_type, FieldType):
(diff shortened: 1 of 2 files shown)


Schamper:
Can you add a unit tests showcasing why this is necessary?

flow/record/fieldtypes/__init__.py
```diff
@@ -283,6 +283,9 @@ def __repr__(self):
         result = str(self)
         return result
 
+    def __hash__(self):
+        return super.__hash__(self)
```
pyrco, line 287:
```suggestion
        return super().__hash__(self)
```

But I think it is better to not depend on the MRO and use the hash of the thing that actually contains the data:

```suggestion
        return _dt.__hash__(self)
```

Poeloe: Changed it to the latter ;)

pyrco (approved):

PR #78, opened 2023-08-03
Make datetime fieldtypes timezone aware

yunzheng:
This PR will add the following:

- Record datetime fields are now offset-aware by default
- Naive datetime fields are converted to UTC
- Support for packing/unpacking aware datetimes

Note that comparing to naive datetime objects will now break and is also in line with default Python behaviour

To ensure uniform datetime field output they are always displayed in UTC
To disable this behaviour you can set the environment variable  `FLOW_RECORD_TZ=NONE`

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -1,22 +1,19 @@
+from __future__ import annotations
+
 import binascii
 import math
 import os
 import pathlib
 import re
 import sys
+import warnings
 from binascii import a2b_hex, b2a_hex
 from datetime import datetime as _dt
 from datetime import timezone
 from posixpath import basename, dirname
 from typing import Any, Optional, Tuple
-from zoneinfo import ZoneInfo
-
-try:
-    import urlparse
-except ImportError:
-    import urllib.parse as urlparse
-
-import warnings
+from urllib.parse import urlparse
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from flow.record.base import FieldType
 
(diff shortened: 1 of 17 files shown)


flow/record/fieldtypes/__init__.py
```diff
@@ -32,6 +40,27 @@
 path_type = pathlib.PurePath
 
 
+def flow_record_tz(default_tz: str = "UTC") -> Optional[ZoneInfo]:
+    """Return a ZoneInfo object based on the ``FLOW_RECORD_TZ`` environment variable.
```
Schamper, line 44:
```suggestion
    """Return a ``ZoneInfo`` object based on the ``FLOW_RECORD_TZ`` environment variable.
```

flow/record/fieldtypes/__init__.py
```diff
@@ -32,6 +40,27 @@
...
+    Args:
+        default_tz: default timezone if ``FLOW_RECORD_TZ`` is not set (default: UTC)
+
+    Returns:
+        None if ``FLOW_RECORD_TZ=NONE`` otherwise ``ZoneInfo(FLOW_RECORD_TZ)``
+    """
+    tz = os.environ.get("FLOW_RECORD_TZ", default_tz)
+    if tz.upper() == "NONE":
+        return None
+    return ZoneInfo(tz)
```
Schamper, line 55: Perhaps throw a try/except on here for `ZoneInfoNotFoundError`? Or would you want it to intentionally throw an error?

Might be useful to show a warning and then default to UTC (straight from `datetime.timezone.utc` to prevent further errors with `ZoneInfo`)? On Windows this will currently always error if you're missing the `tzdata` package.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
 ]
 dependencies = [
     "msgpack>=0.5.2",
+    "backports.zoneinfo[tzdata]; python_version<'3.9'",
```
Schamper, line 27:
```suggestion
    "backports.zoneinfo[tzdata]; python_version<'3.9'",
    "tzdata; platform_system=='Windows'",
```

Schamper (changes requested):

flow/record/fieldtypes/__init__.py
fox-srt: Showing a warning and falling back to UTC is a good idea, will implement this.

yunzheng: Showing a warning and then falling back to UTC is a good idea. will implement this.

Schamper (approved):

PR #81, opened 2023-08-24
Add GitHub workflow to test extra compatibility

yunzheng:
As flow.record still targets Python 3.7 as a minimum, we also test:

- Python 3.7
- Python 3.8
- Windows, Python 3.9+

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -1,0 +1,78 @@
+name: Python Compatibility
+on: [push, pull_request, workflow_dispatch]
+
+jobs: 
+  test-py37-py38:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,78 @@
...
+            tox-env: "py39"
+            allow_failure: false
+          - python-version: "3.10"
+            python-include: "python3.10"
+            tox-env: "py310"
+            allow_failure: false
+          - python-version: "3.11"
+            python-include: "python3.11"
+            tox-env: "py311"
+            allow_failure: false
```
pyrco, line 62: We still allow 3.11 to fail as we keep running into some compatibility issues while 3.11 is stabilizing.

yunzheng: fixed

PR #82, opened 2023-08-25
Remove sed from windows test workflow

yunzheng:


Schamper:
@pyrco IIRC this sed was only important for publishing to PyPI to get rid of the commit hashes, right? So it should be fine to remove it from these tests?

pyrco:
> @pyrco IIRC this sed was only important for publishing to PyPI to get rid of the commit hashes, right? So it should be fine to remove it from these tests?

yes

pyrco (approved):

PR #83, opened 2023-09-14
Support file-like objects in record writer

MaxGroot:
Following https://github.com/fox-it/flow.record/pull/59, this Pull Request adds functionality for adapters of `flow.record` to write to file-like objects.

PR #84, opened 2023-09-14
Feature/gcs adapter

MaxGroot:
Depends on https://github.com/fox-it/flow.record/pull/83. 

This pull request adds a Google Cloud Storage adapter for flow.record. It supports both reading from and writing to Google Cloud Storage in a streaming manner.

### DIFF
--- a/flow/record/adapter/gcs.py
+++ b/flow/record/adapter/gcs.py
@@ -1,0 +1,97 @@
+import logging
+from fnmatch import fnmatch
+from typing import Iterator, Union
+
+from google.cloud.storage.client import Client
+from google.cloud.storage.fileio import BlobReader, BlobWriter
+
+from flow.record.adapter import AbstractReader, AbstractWriter
+from flow.record.base import Record, RecordAdapter
+from flow.record.selector import CompiledSelector, Selector
+
+__usage__ = """
+Google Cloud Storage adapter
+---
+Read usage: rdump gcs://[PROJECT-ID]:[BUCKET-ID]?path=[PATH]
+[PROJECT-ID]: Google Cloud Project ID
+[BUCKET-ID]: Bucket ID
+[path]: Path to look for files, with support for glob-pattern matching
+
+Write usage: rdump gcs://[PROJECT-ID]:[BUCKET-ID]?path=[PATH]
+[PROJECT-ID]: Google Cloud Project ID
+[BUCKET-ID]: Bucket ID
+[path]: Path to write records to
+"""
+
+log = logging.getLogger(__name__)
+
+GLOB_CHARACTERS = "*?[]"
+
+
+class GcsReader(AbstractReader):
+    def __init__(self, uri: str, path: str, selector: Union[None, Selector, CompiledSelector] = None, **kwargs) -> None:
+        self.selector = selector
+        project_name = uri[: uri.find(":")]
+        bucket_name = uri[uri.find(":") + 1 :]
+
+        self.gcs = Client(project=project_name)
+        self.bucket = self.gcs.bucket(bucket_name)
+
+        # GCS Doesn't support iterating blobs using a glob pattern, so we have to do that ourselves.
+        # To split the path prefix from the glob-specific stuff, we have to find the first place where
+        # the glob starts. We'll then go through all files that match the path prefix before the glob,
+        # and do fnmatch ourselves to check whether any given blob matches with our glob.
+        lowest_pos = min([path.find(char) if path.find(char) >= 0 else float("inf") for char in GLOB_CHARACTERS])
+        if lowest_pos == float("inf"):
+            # No glob character was found
+            self.glob = None
+            self.prefix = path
+        else:
+            # Split the glob and the prefix
+            self.prefix = path[:lowest_pos]
+            self.glob = path
+
+    def __iter__(self) -> Iterator[Record]:
+        blobs = self.gcs.list_blobs(bucket_or_name=self.bucket, prefix=self.prefix)
+        for blob in blobs:
+            if blob.size == 0:  # Skip empty files
+                continue
+            if self.glob and not fnmatch(blob.name, self.glob):
+                continue
+            blobreader = BlobReader(blob)
+
+            # Give the file-like object to RecordAdapter so it will select the right adapter by peeking into the stream
+            reader = RecordAdapter(fileobj=blobreader, out=False, selector=self.selector)
+            for record in reader:
+                yield record
+
+    def close(self) -> None:
+        self.gcs.close()
+
+
+class GcsWriter(AbstractWriter):
+    def __init__(self, uri: str, path: str, **kwargs):
+        project_name = uri[: uri.find(":")]
+        bucket_name = uri[uri.find(":") + 1 :]
+        self.writer = None
+
+        self.gcs = Client(project=project_name)
+        self.bucket = self.gcs.bucket(bucket_name)
+
+        blob = self.bucket.blob(path)
+        self.writer = BlobWriter(blob, ignore_flush=True)
+        self.adapter = RecordAdapter(url=path, fileobj=self.writer, out=True, **kwargs)
+
+    def write(self, record: Record) -> None:
+        self.adapter.write(record)
+
+    def flush(self) -> None:
+        # https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.fileio.BlobWriter)
+        # Flushing without closing is not supported by the remote service and therefore calling it on this class
+        # normally results in io.UnsupportedOperation. However, that behavior is incompatible with some consumers and
+        # wrappers of fileobjects in Python.
+        pass
+
+    def close(self) -> None:
+        if self.writer:
+            self.writer.close()
--- a/flow/record/adapter/jsonfile.py
+++ b/flow/record/adapter/jsonfile.py
@@ -23,7 +23,7 @@ class JsonfileWriter(AbstractWriter):
 
     def __init__(self, path, indent=None, descriptors=True, **kwargs):
         self.descriptors = str(descriptors).lower() in ("true", "1")
-        self.fp = record.open_path(path, "w")
+        self.fp = record.open_file(path, "w")
         if isinstance(indent, str):
             indent = int(indent)
         self.packer = JsonRecordPacker(indent=indent, pack_descriptors=self.descriptors)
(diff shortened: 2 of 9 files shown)


flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,97 @@
...
+log = logging.getLogger(__name__)
+
+GLOB_CHARACTERS = "*?[]"
+
+
+class GcsReader(AbstractReader):
+    def __init__(self, uri: str, path: str, selector: Union[None, Selector, CompiledSelector] = None, **kwargs) -> None:
+        self.selector = selector
+        project_name = uri[: uri.find(":")]
+        bucket_name = uri[uri.find(":") + 1 :]
```
yunzheng, line 35:
```suggestion
        project_name, _, bucket_name = uri.partition(":")
```

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.gcs = Client(project=project_name)
+        self.bucket = self.gcs.bucket(bucket_name)
+
+        # GCS Doesn't support iterating blobs using a glob pattern, so we have to do that ourselves.
+        # To split the path prefix from the glob-specific stuff, we have to find the first place where
+        # the glob starts. We'll then go through all files that match the path prefix before the glob,
+        # and do fnmatch ourselves to check whether any given blob matches with our glob.
+        lowest_pos = min([path.find(char) if path.find(char) >= 0 else float("inf") for char in GLOB_CHARACTERS])
```
yunzheng, line 44: i think you can also use `partition()` here to make the code more concise. Also, does this support multiple glob patterns?

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,97 @@
...
+                yield record
+
+    def close(self) -> None:
+        self.gcs.close()
+
+
+class GcsWriter(AbstractWriter):
+    def __init__(self, uri: str, path: str, **kwargs):
+        project_name = uri[: uri.find(":")]
+        bucket_name = uri[uri.find(":") + 1 :]
```
yunzheng, line 75:
```suggestion
        project_name, _, bucket_name = uri.partition(":")
```

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,97 @@
...
+    def flush(self) -> None:
+        # https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.fileio.BlobWriter)
+        # Flushing without closing is not supported by the remote service and therefore calling it on this class
+        # normally results in io.UnsupportedOperation. However, that behavior is incompatible with some consumers and
+        # wrappers of fileobjects in Python.
+        pass
+
+    def close(self) -> None:
+        if self.writer:
+            self.writer.close()
```
yunzheng, line 97:
```suggestion
    def close(self) -> None:
        if self.writer:
            self.writer.close()
            self.writer = None
```

yunzheng (changes requested):

flow/record/adapter/gcs.py
MaxGroot: As far as I could tell, `partition()` isn't really suitable for choosing the first occurence out of multiple seperator characters. To make the code a bit more readable, I opted to feed `min()` an iterator of glob characters and defaulting to `-1` if none matched. 

This PR uses `fnmatch` to verify whether or not the path matches with the pattern, and you can use multiple glob characters in your matching:

```
In [1]: from fnmatch import fnmatch

In [2]: pattern = "*/*/*/t?st.records"

In [3]: fnmatch("a/b/c/tst.records", pattern)
Out[3]: False

In [4]: fnmatch("a/b/c/test.records", pattern)
Out[4]: True
```

Is this what you mean with supporting multiple glob patterns?

Schamper: In `dissect.target` we use a regex for this:
```python
re_glob_index = re.compile(r"(?<=\/)[^\/]*[*?[]\/?")


def glob_split(pattern: str, alt_separator: str = "") -> str:
    # re_glob_index expects a normalized pattern
    pattern = normalize(pattern, alt_separator=alt_separator)

    first_glob = re_glob_index.search(pattern)

    if not first_glob:
        return pattern, ""

    pos = first_glob.start()
    return pattern[:pos], pattern[pos:]
```

yunzheng: You can also use `re.split` in combination with `maxsplit`, for example:

```python
>>> path = "/data/lake/*/citrix/????"
>>> parts = re.split("[\?\*\[\]]", path, maxsplit=1)
>>> print(parts)
['/data/lake/', '/citrix/????']

>>> parts = re.split("[\?\*\[\]]", "/data/lake/test", maxsplit=1)
>>> print(parts)
['/data/lake/test']
```

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,88 @@
...
+from google.cloud.storage.fileio import BlobReader, BlobWriter
+
+from flow.record.adapter import AbstractReader, AbstractWriter
+from flow.record.base import Record, RecordAdapter
+from flow.record.selector import CompiledSelector, Selector
+
+__usage__ = """
+Google Cloud Storage adapter
+---
+Read usage: rdump gcs://[PROJECT-ID]:[BUCKET-ID]?path=[PATH]
```
yunzheng, line 16: Could you change `BUCKET-ID` to `BUCKET_ID` and same for `PROJECT_ID`?

Also, what are your thoughts of making the `gcs://` uri more `gsutil` like compatible?

For example. getting rid of the `path` parameter and access it like:

`gcs://bucket-id/path/`

And an example with specifying the `project-id`:

`gcs://my-fancy-bucket/path/to/records/?project_id=my-project-12345`

yunzheng: This also has the added benefit that you could just access it like this:

```python
with RecordReader("gcs://my-fancy-bucket/records/*.avro", project_id="my-project-12345") as reader:
    for record in reader:
        print(record)

File: flow/record/adapter/gcs.py [hunk 6]
```diff
@@ -0,0 +1,88 @@
+import logging
+import re
+from fnmatch import fnmatch
+from typing import Iterator, Union
+
+from google.cloud.storage.client import Client
+from google.cloud.storage.fileio import BlobReader, BlobWriter
```
pyrco, line 7: This will pull in some major external dependencies. Maybe good to hide it behind a conditional import and add the dependencies as extras?

pyrco (changes requested):

flow/record/adapter/gcs.py
Schamper: Adapters are lazily loaded and it's not added as a dependency in pyproject.toml, so it should be fine? Although we could do with an extra `flow.record[gcs]` which does add these dependencies.

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,87 @@
...
+Read usage: rdump gcs://[BUCKET_ID]/path?project_id=[PROJECT_ID]
+[PROJECT_ID]: Google Cloud Project ID
+[BUCKET_ID]: Bucket ID
+[path]: Path to look for files, with support for glob-pattern matching
+
+Write usage: rdump gcs://[BUCKET_ID]/path?project_id=[PROJECT_ID]
+[PROJECT_ID]: Google Cloud Project ID
+[BUCKET_ID]: Bucket ID
+[path]: Path to write records to
+"""
```
yunzheng, line 25: Made the usage string a bit more concise, and also changed `project_id` -> `project` and make it optional.

```suggestion
__usage__ = """
Google Cloud Storage adapter
---
Read usage: rdump gcs://[BUCKET_ID]/path?project=[PROJECT]
Write usage: rdump -w gcs://[BUCKET_ID]/path?project=[PROJECT]

[BUCKET_ID]: Bucket ID
[path]: Path to read from or write to, supports glob-pattern matching when reading

Optional arguments:
    [PROJECT]: Google Cloud Project ID, If not passed, falls back to the default inferred from the environment.
"""
```

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,87 @@
+import logging
+import re
```
yunzheng, line 2: So we can use some more modern type hinting.

```suggestion
from __future__ import annotations

import logging
import re
```

yunzheng (changes requested):
Tested and it works very nicely. At first I didn't install `fastavro` so when reading some *.avro files it threw unhelpful messages..

So, I wonder if we need to make the file magic detection just raise errors instead of failing silently. Anyway, that is for future reference and not this comment.

Other than that, I have some small suggestions and then it looks good to go!

flow/record/adapter/gcs.py
```diff
@@ -0,0 +1,87 @@
...
+[path]: Path to write records to
+"""
+
+log = logging.getLogger(__name__)
+
+GLOB_CHARACTERS_RE = r"[\[\]\*\?]"
+
+
+class GcsReader(AbstractReader):
+    def __init__(self, uri: str, project: str, selector: Union[None, Selector, CompiledSelector] = None, **kwargs):
```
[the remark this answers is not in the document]
Schamper: If you're going to use | for the Union, also use it for the other Unions 😄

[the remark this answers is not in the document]
yunzheng: > If you're going to use | for the Union, also use it for the other Unions 😄

Oops, updated the code suggestion, also CompiledSelector subclasses from Selector, so we can just use that as the type.

yunzheng (changes requested):
I have some problems writing a gzipped file to a GCS bucket

MaxGroot:
@yunzheng To support transparent compression when writing to filelike objects, I've added `wrap_in_compression` to `base.py`.  This wraps fp in a compressor if this seems sensible based on a given path. This is copied from `open_path`, though changed to always instantiate the compressor using a filelike object. This should solve the problem of writing a gzipped file to a GCS bucket, though cannot test it for realsies. 

An unfortunate consequence is that the `RecordAdapter` method has become just a tad more complex than it already was. We've discussed refactoring it in the past, but I think it will remain complex due to how flexible it is expected to be. It has to account for a URL, a file-like object, and whether or not the adapter should be a writer or a reader, with some complexity mixed in for handling stdio. If you have ideas for refactoring it properly I'm all ears, though that may be something for a different PR.

PR #85, opened 2023-09-15
Add support for Splunk HTTP Event Collector

MaxGroot:
This pull request adds support for the Splunk HTTP Event Collector (HEC) to the Splunk adapter. This is especially useful in scenarios where the already supported TCP Input is not available, for example in Splunk Cloud environments.

Sending data over HTTP Event Collector supports both sending the data in `key=value` format (like the TCP Data Input already uses) and a `json` format. 

Something I'm still unsure of is whether to include the 'builtin' Record Fields such as `_generated`, `_source`, `_classification` and `_version`. Currently, the keyvalue format does _not_ send these fields, but the `json` format does. This difference is caused by the usage of the `JsonRecordPacker` which includes these fields. I don't think it should be inconsistent between the two formats, but what is desired: including or omitting these builtin fields?

### DIFF
--- a/flow/record/adapter/splunk.py
+++ b/flow/record/adapter/splunk.py
@@ -8,13 +8,11 @@ from typing import Optional
 from urllib.parse import urlparse
 
 try:
-    import requests
-    import urllib3
+    import httpx
 
-    urllib3.disable_warnings()
-    HAS_REQUESTS = True
+    HAS_HTTPX = True
 except ImportError:
-    HAS_REQUESTS = False
+    HAS_HTTPX = False
 
 from flow.record.adapter import AbstractReader, AbstractWriter
 from flow.record.base import Record
@@ -187,8 +185,8 @@ class SplunkWriter(AbstractWriter):
             self.sock.connect((self.host, self.port))
             self._send = self._send_tcp
         elif self.protocol in (Protocol.HTTP, Protocol.HTTPS):
-            if not HAS_REQUESTS:
-                raise ImportError("The requests library is required for sending data over HTTP(S)")
+            if not HAS_HTTPX:
+                raise ImportError("The httpx library is required for sending data over HTTP(S)")
 
             scheme = self.protocol.value
             self.token = token
@@ -212,9 +210,8 @@ class SplunkWriter(AbstractWriter):
                 "X-Splunk-Request-Channel": str(uuid.uuid4()),
             }
 
-            self.session = requests.Session()
-            self.session.verify = self.verify
-            self.session.headers.update(self.headers)
+            self.session = httpx.Client(verify=self.verify, headers=self.headers)
+
             self._send = self._send_http
 
     def _cache_records_for_http(self, data: Optional[bytes] = None, flush: bool = False) -> Optional[bytes]:
--- a/flow/record/adapter/splunk.py
+++ b/flow/record/adapter/splunk.py
@@ -66,14 +66,18 @@ class SourceType(Enum):
 def splunkify_key_value(record: Record, tag: Optional[str] = None) -> str:
     ret = []
 
-    ret.append(f'type="{record._desc.name}"')
+    ret.append(f'rdtype="{record._desc.name}"')
 
     if tag is None:
         ret.append("rdtag=None")
     else:
         ret.append(f'rdtag="{tag}"')
 
-    for field in record._desc.fields:
+    for field in record._desc.get_all_fields():
+        # Omit the _version field as the Splunk adapter has no reader support for deserialising records back.
+        if field == "_version":
+            continue
+
         val = getattr(record, field)
         if val is None:
             ret.append(f"{field}=None")
@@ -90,24 +94,24 @@ def splunkify_key_value(record: Record, tag: Optional[str] = None) -> str:
 def splunkify_json(packer: JsonRecordPacker, record: Record, tag: Optional[str] = None) -> str:
     ret = {}
 
-    indexer_fields = {
+    indexer_fields = [
         ("host", "host"),
         ("host", "hostname"),
         ("time", "ts"),
         ("source", "_source"),
-    }
+    ]
 
     # When converting a record to json text for splunk, we distinguish between the 'event' (containing the data) and a
     # few other fields that are splunk-specific for indexing. We add those 'indexer_fields' to the return object first.
-    for dest_field, source_field in indexer_fields:
-        if hasattr(record, source_field):
-            val = getattr(record, source_field)
+    for splunk_name, field_name in indexer_fields:
+        if hasattr(record, field_name):
+            val = getattr(record, field_name)
             if val:
                 if isinstance(val, datetime):
                     # Convert datetime objects to epoch timestamp for reserved fields.
-                    ret[dest_field] = val.timestamp()
+                    ret[splunk_name] = val.timestamp()
                     continue
-                ret[dest_field] = to_str(val)
+                ret[splunk_name] = to_str(val)
 
     record_as_dict = packer.pack_obj(record)
 
@@ -121,9 +125,11 @@ def splunkify_json(packer: JsonRecordPacker, record: Record, tag: Optional[str] 
         del record_as_dict[field]
         record_as_dict[new_field] = val
 
-    # almost done, just have to add the tag and the type (i.e the record descriptor's name) to the event.
+    # Almost done, just have to add the tag and the type (i.e the record descriptor's name) to the event.
     record_as_dict["rdtag"] = tag
-    record_as_dict["type"] = record._desc.name
+
+    # Yes.
+    record_as_dict["rdtype"] = record._desc.name
 
     ret["event"] = record_as_dict
     return json.dumps(ret, default=packer.pack_obj)
@@ -141,6 +147,7 @@ class SplunkWriter(AbstractWriter):
         ssl_verify: bool = True,
         **kwargs,
     ):
+        # If the writer is initiated without a protocol, we assume we will be writing over tcp
         if "://" not in uri:
             uri = f"tcp://{uri}"
 
@@ -174,35 +181,34 @@ class SplunkWriter(AbstractWriter):
         if self.protocol == Protocol.TCP:
             self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.SOL_TCP)
             self.sock.connect((self.host, self.port))
+            return
 
-        elif self.protocol in [Protocol.HTTP, Protocol.HTTPS]:
-            if not HAS_REQUESTS:
-                raise ImportError("The requests library is required for sending data over HTTP(S).")
+        # Protocol is not TCP, so it is either HTTP or HTTPS
+        scheme = self.protocol.value
 
-            self.token = token
+        if not HAS_REQUESTS:
+            raise ImportError("The requests library is required for sending data over HTTP(S).")
 
-            # Assume verify=True unless specified otherwise.
-            self.verify = not (str(ssl_verify).lower() in ("0", "false"))
+        self.token = token
+        if not self.token:
+            raise ValueError("An authorization token is required for the HTTP collector.")
+        if not self.token.startswith("Splunk "):
+            self.token = "Splunk " + self.token
 
-            scheme = self.protocol.value
-            endpoint = "event" if self.sourcetype != SourceType.RECORDS else "raw"
-            port = f":{self.port}" if self.port else ""
-            self.url = f"{scheme}://{self.host}{port}/services/collector/{endpoint}?auto_extract_timestamp=true"
+        # Assume verify=True unless specified otherwise.
+        self.verify = str(ssl_verify).lower() not in ("0", "false")
+        if not self.verify:
+            log.warning("Certification verification is disabled.")
 
-            if not self.token:
-                raise ValueError("An authorization token is required for the HTTP collector.")
+        endpoint = "event" if self.sourcetype != SourceType.RECORDS else "raw"
+        port = f":{self.port}" if self.port else ""
+        self.url = f"{scheme}://{self.host}{port}/services/collector/{endpoint}?auto_extract_timestamp=true"
 
-            if not self.token.startswith("Splunk "):
-                self.token = "Splunk " + self.token
-
-            if not self.verify:
-                log.warning("Certification verification is disabled.")
-
-            self.headers = {
-                "Authorization": self.token,
-                # A randomized value so that Splunk can loadbalance between different incoming datastreams
-                "X-Splunk-Request-Channel": str(uuid.uuid4()),
-            }
+        self.headers = {
+            "Authorization": self.token,
+            # A randomized value so that Splunk can loadbalance between different incoming datastreams
+            "X-Splunk-Request-Channel": str(uuid.uuid4()),
+        }
 
     def _cache_records_for_http(self, data: Optional[bytes] = None, flush: bool = False) -> Optional[bytes]:
         # It's possible to call this function without any data, purely to flush. Hence this check.
@@ -225,7 +231,7 @@ class SplunkWriter(AbstractWriter):
             return
         response = requests.post(self.url, headers=self.headers, verify=self.verify, data=buf)
         if response.status_code != 200:
-            raise Exception(f"{response.text} ({response.status_code})")
+            raise ConnectionError(f"{response.text} ({response.status_code})")
 
     def _send_tcp(self, data: bytes) -> None:
         self.sock.sendall(data)
--- a/flow/record/adapter/splunk.py
+++ b/flow/record/adapter/splunk.py
@@ -144,6 +144,7 @@ def splunkify_json(packer: JsonRecordPacker, record: Record, tag: Optional[str] 
 
 class SplunkWriter(AbstractWriter):
     sock = None
+    session = None
 
     def __init__(
         self,
@@ -160,18 +161,14 @@ class SplunkWriter(AbstractWriter):
 
         if sourcetype is None:
             log.warning("No sourcetype provided, assuming 'records' sourcetype")
-            sourcetype = SourceType.RECORDS.value
+            sourcetype = SourceType.RECORDS
 
         parsed_url = urlparse(uri)
         url_scheme = parsed_url.scheme.lower()
 
-        self.protocol = next((protocol for protocol in Protocol if protocol.value == url_scheme), None)
-        self.sourcetype = next((source for source in SourceType if source.value == sourcetype), None)
+        self.sourcetype = SourceType(sourcetype)
+        self.protocol = Protocol(url_scheme)
 
-        if not self.sourcetype:
-            raise ValueError(f"Unsupported source type {sourcetype}")
-        if not self.protocol:
-            raise ValueError(f"Unsupported protocol {url_scheme}")
         if self.protocol == Protocol.TCP and self.sourcetype != SourceType.RECORDS:
             raise ValueError("For sending data to Splunk over TCP, only the 'records' sourcetype is allowed")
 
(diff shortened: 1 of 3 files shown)


flow/record/adapter/splunk.py
```diff
@@ -1,19 +1,42 @@
...
 [IP]:[PORT]: ip and port to a splunk instance
 [TAG]: optional value to add as "rdtag" output field when writing
+[TOKEN]: Authentication token for sending data over HTTP(S)
+[SOURCETYPE]: Set sourcetype of data. Defaults to records, but can also be set to JSON.
+[SSL_VERIFY]: Whether to verify the server certificate when sending data over HTTP(S). Defaults to True.
 """
 
 log = logging.getLogger(__package__)
 
+# Amount of records to bindle into a single request when sending data over HTTP(S).
```
Schamper, line 37:
```suggestion
# Amount of records to bundle into a single request when sending data over HTTP(S)
```

flow/record/adapter/splunk.py
```diff
@@ -27,7 +50,18 @@
 )
 
 
-def splunkify(record, tag=None):
+class SPLUNK_PROTOCOLS(Enum):
```
Schamper, line 53:
```suggestion
class Protocol(Enum):
```
Unless there's a specific reason to prefix it with `Splunk`, I believe it being in this file should be context enough. Also enum classes are usually singular and Camelcased, with the values UPPER_CASE.

flow/record/adapter/splunk.py
```diff
@@ -27,7 +50,18 @@
...
+    HTTP = "http"
+    HTTPS = "https"
+    TCP = "tcp"
+
+
+class SPLUNK_SOURCETYPES(Enum):
```
Schamper, line 59:
```suggestion
class SourceType(Enum):
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +85,161 @@ def splunkify(record, tag=None):
...
+        if self.protocol == SPLUNK_PROTOCOLS.TCP and self.sourcetype != SPLUNK_SOURCETYPES.RECORDS:
+            raise ValueError("For sending data to splunk over TCP, only the 'records' sourcetype is allowed.")
+
+        self.host = parsed_url.hostname
+        self.port = parsed_url.port
         self.tag = tag
-        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.SOL_TCP)
-        self.sock.connect((host, port))
-        self.descriptors = {}
+        self.record_buffer = set()
```
Schamper, line 154: Using a set will make the buffer entries unique, is that intended?

Schamper (changes requested):

flow/record/adapter/splunk.py
```diff
@@ -51,38 +85,161 @@ def splunkify(record, tag=None):
...
+    record_as_dict = packer.pack_obj(record)
+    record_as_dict["rdtag"] = tag
+    for dest_field, source_field in indexer_fields:
+        if hasattr(record, source_field):
+            val = getattr(record, source_field)
+            if val:
+                if isinstance(val, datetime):
+                    # Convert datetime objects to epoch timestamp for reserved fields.
+                    epoch = (val - datetime(1970, 1, 1, tzinfo=timezone.utc)).total_seconds()
+                    ret[dest_field] = epoch
```
yunzheng, line 106:
```suggestion
                    ret[dest_field] = val.timestamp()
```

flow/record/adapter/splunk.py
```diff
@@ -1,19 +1,42 @@
...
 from flow.record.adapter import AbstractReader, AbstractWriter
+from flow.record.base import Record
+from flow.record.jsonpacker import JsonRecordPacker
 from flow.record.utils import to_base64, to_bytes, to_str
 
 __usage__ = """
 Splunk output adapter (writer only)
 ---
-Write usage: rdump -w splunk://[IP]:[PORT]?tag=[TAG]
+Write usage: rdump -w splunk+[PROTOCOL]://[IP]:[PORT]?tag=[TAG]&token=[TOKEN]&sourcetype=[SOURCETYPE]
```
yunzheng, line 27: `[PROTOCOL]` is not documented here, i assume if it's not defined it will default to TCP.

yunzheng (changes requested):

flow/record/adapter/splunk.py
```diff
@@ -51,38 +85,161 @@ def splunkify(record, tag=None):
...
+            return
+        buf = b""
+        for record_data in self.record_buffer:
+            buf += record_data + b"\n"
+
+        if not len(buf):
+            return
+
+        # Remove the last newline as it is duplicate.
+        buf = buf[:-1]
```
yunzheng, line 201:
```suggestion
        buf = b"\n".join(self.record_buffer)

        if not len(buf):
            return
```

A set is unordered, not a big deal here but could be annoying when debugging.
Are you using a set to deduplicate records? if not i would just use a list.

Also, consider moving the buffering code outside this method.

MaxGroot:
While working on the HTTP adapter it came to mind that there seems to be undesired behavior in the current Splunk adapter whenever a record has a field with the name `type`, as `type` is currently being used by the splunk adapter to denote the `name` property of the record's `RecordDescriptor`. For example,

```
In [1]: from flow.record import RecordDescriptor, RecordWriter
In [2]: descriptor = RecordDescriptor("my/record", [("string", "type")])
In [3]: record = descriptor(type="some_value")
In [4]: record
Out[4]: <my/record type='some_value'
In [6]: writer = RecordWriter("splunk://localhost:4444")
In [7]: writer.write(record)
 ```
 
 This will currently yield the following input for Splunk:
```
netcat -lvp 4444
Listening on 0.0.0.0 4444
Connection received on localhost 1337
type="my/record" rdtag=None type="some_value"
``` 

As you can see, the `type` field is set twice, for two very different things.
I think the Splunk adapter should to the same thing for `type` as it currently does for `tag`. The 'meta' field should be prefixed with `rd`, which would yield the following input in Splunk:

```
rdtype="my/record" rdtag=None type="some_value"
```
This however might break existing Splunk dashboards that build on top of disssect. In my experience records from dissect rarely have a 'type' field, which might explain why things haven't broken down in the past. Having said that, I would still consider the above (current) behavior to be a bug, and the introduction of the HTTP collector support of this PR requires me to either also implement the behavior in this PR, or fix it now.

Do you think the record descriptor name should be sent as `rdtype` rather than `type`? 

And if so, should it be implemented in a separate PR first, or should I implement it within this PR?

**Comment by Schamper** [comment]
> While working on the HTTP adapter it came to mind that there seems to be undesired behavior in the current Splunk adapter whenever a record has a field with the name `type`, as `type` is currently being used by the splunk adapter to denote the `name` property of the record's `RecordDescriptor`. For example,
> 
> ```
> In [1]: from flow.record import RecordDescriptor, RecordWriter
> In [2]: descriptor = RecordDescriptor("my/record", [("string", "type")])
> In [3]: record = descriptor(type="some_value")
> In [4]: record
> Out[4]: <my/record type='some_value'
> In [6]: writer = RecordWriter("splunk://localhost:4444")
> In [7]: writer.write(record)
> ```
> 
> This will currently yield the following input for Splunk:
> 
> ```
> netcat -lvp 4444
> Listening on 0.0.0.0 4444
> Connection received on localhost 1337
> type="my/record" rdtag=None type="some_value"
> ```
> 
> As you can see, the `type` field is set twice, for two very different things. I think the Splunk adapter should to the same thing for `type` as it currently does for `tag`. The 'meta' field should be prefixed with `rd`, which would yield the following input in Splunk:
> 
> ```
> rdtype="my/record" rdtag=None type="some_value"
> ```
> 
> This however might break existing Splunk dashboards that build on top of disssect. In my experience records from dissect rarely have a 'type' field, which might explain why things haven't broken down in the past. Having said that, I would still consider the above (current) behavior to be a bug, and the introduction of the HTTP collector support of this PR requires me to either also implement the behavior in this PR, or fix it now.
> 
> Do you think the record descriptor name should be sent as `rdtype` rather than `type`?
> 
> And if so, should it be implemented in a separate PR first, or should I implement it within this PR?

Yes.

**Comment by yunzheng** [comment]
> Something I'm still unsure of is whether to include the 'builtin' Record Fields such as _generated, _source, _classification and _version. Currently, the keyvalue format does not send these fields, but the json format does. This difference is caused by the usage of the JsonRecordPacker which includes these fields. I don't think it should be inconsistent between the two formats, but what is desired: including or omitting these builtin fields?

I would say `_version` is useless as the Splunk adapter has no reader support for deserialising it back using its original RecordDescriptor.

`_generated` could be useful to have imo.

`_source` can also be useful if people are using it. This is meant for extra metadata, for example you could put the disk image filename or path here which can be useful to determine where the record originated from.

`_classification` can also be useful if people are using it :) - eg: TLP:AMBER, TLP:WHITE etc.

File: flow/record/adapter/splunk.py [hunk 8]
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+                    ret[splunk_name] = val.timestamp()
+                    continue
+                ret[splunk_name] = to_str(val)
+
+    record_as_dict = packer.pack_obj(record)
+
+    # These fields end up in the 'event', but we have a few reserved field names. If those field names are in the
+    # record, we prefix them with 'rd_' (short for record descriptor)
+    for field in RESERVED_SPLUNK_FIELDS:
+        if field not in record_as_dict.keys():
```
Schamper, line 121:
```suggestion
        if field not in record_as_dict:
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+            continue
+        new_field = f"rd_{field}"
+        val = record_as_dict[field]
+        del record_as_dict[field]
+        record_as_dict[new_field] = val
```
Schamper, line 126:
```suggestion
        record_as_dict[new_field] = record_as_dict[field]
        del record_as_dict[field]
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+        sourcetype: Optional[str] = None,
+        ssl_verify: bool = True,
+        **kwargs,
+    ):
+        # If the writer is initiated without a protocol, we assume we will be writing over tcp
+        if "://" not in uri:
+            uri = f"tcp://{uri}"
+
+        if sourcetype is None:
+            log.warning("No sourcetype provided, assuming 'records' sourcetype.")
```
Schamper, line 155:
```suggestion
            log.warning("No sourcetype provided, assuming 'records' sourcetype")
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+            sourcetype = SourceType.RECORDS.value
+
+        parsed_url = urlparse(uri)
+        url_scheme = parsed_url.scheme.lower()
+
+        self.protocol = next((protocol for protocol in Protocol if protocol.value == url_scheme), None)
+        self.sourcetype = next((source for source in SourceType if source.value == sourcetype), None)
 
+        if not self.sourcetype:
+            raise ValueError(f"Unsupported source type {sourcetype}.")
```
Schamper, line 165:
```suggestion
            raise ValueError(f"Unsupported source type {sourcetype}")
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+        if not self.protocol:
+            raise ValueError(f"Unsupported protocol {url_scheme}.")
```
Schamper, line 167:
```suggestion
            raise ValueError(f"Unsupported protocol {url_scheme}")
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+        if self.protocol == Protocol.TCP and self.sourcetype != SourceType.RECORDS:
+            raise ValueError("For sending data to splunk over TCP, only the 'records' sourcetype is allowed.")
```
Schamper, line 169:
```suggestion
            raise ValueError("For sending data to Splunk over TCP, only the 'records' sourcetype is allowed")
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+        if self.protocol == Protocol.TCP:
+            self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.SOL_TCP)
+            self.sock.connect((self.host, self.port))
+            return
+
+        # Protocol is not TCP, so it is either HTTP or HTTPS
+        scheme = self.protocol.value
+
+        if not HAS_REQUESTS:
+            raise ImportError("The requests library is required for sending data over HTTP(S).")
```
Schamper, line 190:
```suggestion
            raise ImportError("The requests library is required for sending data over HTTP(S)")
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+
+        self.token = token
+        if not self.token:
+            raise ValueError("An authorization token is required for the HTTP collector.")
```
Schamper, line 194:
```suggestion
            raise ValueError("An authorization token is required for the HTTP collector")
```

flow/record/adapter/splunk.py
```diff
@@ -51,38 +91,183 @@ def splunkify(record, tag=None):
...
+        if not self.token.startswith("Splunk "):
+            self.token = "Splunk " + self.token
+
+        # Assume verify=True unless specified otherwise.
+        self.verify = str(ssl_verify).lower() not in ("0", "false")
+        if not self.verify:
+            log.warning("Certification verification is disabled.")
```
Schamper, line 201:
```suggestion
            log.warning("Certificate verification is disabled")
```

flow/record/adapter/splunk.py
```diff
@@ -23,21 +47,37 @@
...
 
     if tag is None:
         ret.append("rdtag=None")
     else:
         ret.append(f'rdtag="{tag}"')
 
-    for field in record._desc.fields:
+    for field in record._desc.get_all_fields():
+        # Omit the _version field as the Splunk adapter has no reader support for deserialising records back.
+        if field == "_version":
```
Schamper, line 78: You don't do this in the JSON serialisation, is that intentional? And do the other `_` prefixed fields work fine?

Schamper (changes requested):

flow/record/adapter/splunk.py
MaxGroot: > You don't do this in the JSON serialisation, is that intentional? And do the other `_` prefixed fields not break anything? I vaguely remember being told not to do that, but I don't remember the reason.

Yes.

MaxGroot: Jokes aside, you were right that the JSON and records sourcetype were inconsistent in whether to include `_version`, it is now omitted for both. 

If I recall correctly, the `_` prefix should be fine as the reserved fields are guarded under the `SPLUNK_RESERVED_FIELDS` constant. I also vaguely recall warnings about breaking things, maybe you can ask Splunk people on your end? I haven't really touched anything Splunk related after the "Yes." incident.

<sup><sub>I miss you so much it hurts sometimes</sub></sup>

Schamper: I asked our Splunk experts internally and they recommended to add a custom prefix to `_` prefixed fields. This recommendation from Splunk can be found on this page: https://docs.splunk.com/Documentation/Splunk/7.3.1/Data/Configureindex-timefieldextraction (maybe nice to include this URL as a comment).

Schamper (changes requested):

flow/record/adapter/splunk.py
MaxGroot: I added a `rd_` prefix to the `_` prefixed fields (`_generated`, `_classification`), but while doing so ran into ambiguity for the `_source` field.  Sending data to the HEC supports overriding the `source` field while feeding data, and [Splunk's definition](https://docs.splunk.com/Splexicon:Source) of `source` seems to map perfectly on the usage of `_source` in flow.record. Therefore, the HEC writer already used `_source` for setting the `source` property.

However, the TCP forwarding method currently ignores `_source` in its entirety. No `source` value is set, and if a record has a field named `source` (e.g. `unix/user` from dissect.target), it is changed to `rd_source`. 

I think using `_source` field for Splunk's `source` would be nice, when available. Therefore I've added this to the TCP writer in the last commit and added a test to show how it works. One problem: I don't have access to the dissect technical addon for which the adapter was originally written, so I cannot test if this actually works.

These Splunk experts you speak of, do they have a Github account by any chance? It would be nice if we could wrap up this PR..

Schamper: I couldn't get them to comment on GitHub directly so you'll have to deal with me for a little longer. I hope you don't mind, otherwise we can talk it out over a nice dinner. 

I've relayed your proposal and come back with the conclusion that a) it's nice information to have available in Splunk, but b) that the `source` field is not the correct field for this. At least in the Dissect Splunk Technical Addon, the record descriptor type is used as `source` field, and that is a more appropriate use of that field.

We've come up with the following proposal (record field -> Splunk field):
- `source` -> `rd_source`
- `_source` -> `rd__source`

If you agree, it might be nice to also apply this to the other `_`-prefixed fields, where you currently remove the extra underscore. That way it provides a clear distinction between what fields were originally "private" flow record fields.

MaxGroot: Thanks for relaying, I've implemented the proposal. 

What's for dessert?

Schamper: Thanks, that looks good.

Dessert is all about the experience. It's not just what's on the plate but who's across from you. What about later, maybe we skip right to the sequel of what we finished last time? I promise it's a blockbuster.

flow/record/adapter/splunk.py
```diff
@@ -1,88 +1,280 @@
...
+        del record_as_dict[field]
+
+    # Almost done, just have to add the tag and the type (i.e the record descriptor's name) to the event.
+    record_as_dict["rdtag"] = tag
+
+    # Yes.
+    record_as_dict["rdtype"] = record._desc.name
+
+    ret["event"] = record_as_dict
+    return json.dumps(ret, default=packer.pack_obj)
```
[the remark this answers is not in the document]
MaxGroot: Turns out that the `JsonRecordPacker` leaves record values of type `posix_path` as-is. Because of that, `record_as_dict` will contain non-serializable fields. When `json.dumps` attempts to serialize a value of type `posix_path`, it will cal the `JsonRecordPacker` on that individual value, whereas earlier it was called on _a record containing said value_. `JsonRecordPacker` will then convert it to a string.

Meaning, leaving default out here will break on records containing `posix_path` field types. We could look into fixing this in the `JsonRecordPacker`, or we could drop everything and go home, seeing as it's 5PM and it's a Friday.

Schamper (changes requested):

flow/record/adapter/splunk.py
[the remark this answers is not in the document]
Schamper: Ah yeah that looks like an oversight when adding the `path` type, that should've been added to the JSON packer as well. Let's fix that in another PR.

See you at home.

flow/record/adapter/splunk.py
```diff
@@ -1,89 +1,284 @@
...
 import logging
 import socket
+import uuid
+from datetime import datetime
+from enum import Enum
+from typing import Optional
+from urllib.parse import urlparse
+
+try:
+    import requests
```
Schamper, line 11: Would you mind changing out to `httpx`? Should be a very easy change.

Schamper (changes requested):

flow/record/adapter/splunk.py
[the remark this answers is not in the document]
yunzheng: > Ah yeah that looks like an oversight when adding the `path` type, that should've been added to the JSON packer as well. Let's fix that in another PR.
> 
> See you at home.

Was this eventually fixed in another PR? If not, please make an issue.

Schamper (approved):

PR #86, opened 2023-10-09
Add behaviour to always use datetime.UTC if there is no zoneinfo available

Miauwkeru:
DIS-2238

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -61,7 +61,8 @@ def flow_record_tz(*, default_tz: str = "UTC") -> Optional[ZoneInfo | UTC]:
         return None
 
     if not HAS_ZONE_INFO:
-        warnings.warn("Cannot use FLOW_RECORD_TZ due to missing zoneinfo module, defaulting to 'UTC'.")
+        if tz != "UTC":
+            warnings.warn("Cannot use FLOW_RECORD_TZ due to missing zoneinfo module, defaulting to 'UTC'.")
         return UTC
 
     try:


flow/record/fieldtypes/__init__.py
```diff
@@ -15,9 +15,15 @@
...
 try:
-    from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+    try:
+        from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+    except ImportError:
+        from backports.zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+    HAS_ZONE_INFO = True
 except ImportError:
-    from backports.zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+    warnings.waring("Could not use zoneinfo, using the default timezone 'UTC'.")
```
yunzheng, line 24:
```suggestion
    warnings.warn("Could not use zoneinfo, using the default timezone 'UTC'.")
```

yunzheng: typo here, but I would remove the warning here all together, and move it to my suggestion below.
I think raising the warning only if FLOW_RECORD_TZ has been set is the more user friendly.

flow/record/fieldtypes/__init__.py
```diff
@@ -50,9 +56,13 @@ def flow_record_tz(*, default_tz: str = "UTC") -> Optional[ZoneInfo | UTC]:
     Returns:
         None if ``FLOW_RECORD_TZ=NONE`` otherwise ``ZoneInfo(FLOW_RECORD_TZ)`` or ``UTC`` if ZoneInfo is not found.
     """
+    if not HAS_ZONE_INFO:
+        return UTC
+
     tz = os.environ.get("FLOW_RECORD_TZ", default_tz)
     if tz.upper() == "NONE":
         return None
+
```
yunzheng, line 65: Can you insert something like this here:

```python
if not HAS_ZONE_INFO and tz != "UTC":
    warning.warn("Cannot use FLOW_RECORD_TZ due to missing zoneinfo module, defaulting to 'UTC'.")
    return UTC
```

yunzheng (changes requested):

flow/record/fieldtypes/__init__.py
yunzheng: Good catch, then something like this?

```python
if not HAS_ZONE_INFO:
    if tz != "UTC":
        warning.warn("Cannot use FLOW_RECORD_TZ due to missing zoneinfo module, defaulting to 'UTC'.")
    return UTC
```

So always return UTC when zoneinfo is not available, and only print the warning if the user actually configured FLOW_RECORD_TZ to something non UTC.

flow/record/fieldtypes/__init__.py
```diff
@@ -50,9 +55,15 @@ def flow_record_tz(*, default_tz: str = "UTC") -> Optional[ZoneInfo | UTC]:
     Returns:
         None if ``FLOW_RECORD_TZ=NONE`` otherwise ``ZoneInfo(FLOW_RECORD_TZ)`` or ``UTC`` if ZoneInfo is not found.
     """
+
     tz = os.environ.get("FLOW_RECORD_TZ", default_tz)
     if tz.upper() == "NONE":
         return None
+
+    if not HAS_ZONE_INFO:
+        warnings.warn("Cannot use FLOW_RECORD_TZ due to missing zoneinfo module, defaulting to 'UTC'.")
```
yunzheng, line 64: No warning needs to be given in these cases:

* FLOW_RECORD_TZ is not set, it will default to "UTC"
* FLOW_RECORD_TZ is set to "UTC"

In those cases it already returns UTC, so no warnings needs to be given. This ensures that in default behaviour no unnecessary warnings are triggered.

```suggestion
        if tz != "UTC":
            warnings.warn("Cannot use FLOW_RECORD_TZ due to missing zoneinfo module, defaulting to 'UTC'.")
```
Note `tz` here is a string from the environment variable `FLOW_RECORD_TZ` and not a timezone object

yunzheng (approved):

yunzheng:
LGTM

Schamper (approved):

PR #88, opened 2023-10-11
Drop Python 3.7 support

yunzheng:
Python 3.7 is EOL since 27 June 2023.
Minimal supported Python version for flow.record is now 3.8.

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -38,35 +38,3 @@ jobs:
         with:
           name: coverage
           path: coverage.xml
-  test-windows:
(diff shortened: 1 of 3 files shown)


yunzheng:
Looks like `_PosixPath` is removed in Python 3.12

pyrco:
> Looks like `_PosixPath` is removed in Python 3.12

Yes, 3.12 is expected to fail for now on this issue.

Regarding the removal of the local windows tests, you can now also revert the change to tox.ini that was made when these  tests were added.

pyrco (approved):

PR #92, opened 2023-10-24
Update `elastic.py` adapter

0xbart:
Update `elastic.py` adapter, fix AttributeError if invalid uri is given, add `verify_certs` flag to optional arguments.
Also check with `hasattr` if 'self.es' is set. This can be None, for example if invalid uri (such as missing port number) is given.

### DIFF
--- a/flow/record/adapter/elastic.py
+++ b/flow/record/adapter/elastic.py
@@ -1,6 +1,7 @@
 import logging
 import queue
 import threading
+import hashlib
 from typing import Iterator, Union
 
 import elasticsearch
@@ -18,12 +19,12 @@ ElasticSearch adapter
 Write usage: rdump -w elastic+[PROTOCOL]://[IP]:[PORT]?index=[INDEX]
 Read usage: rdump elastic+[PROTOCOL]://[IP]:[PORT]?index=[INDEX]
 [IP]:[PORT]: ip and port to elastic host
-[INDEX]: index to write to or read from
 [PROTOCOL]: http or https. Defaults to https when "+[PROTOCOL]" is omitted
 
 Optional arguments:
-  [INDEX]: name of the index to use
-  [VERIFY_CERTS]: verify certs of Elasticsearch instance
+  [INDEX]: name of the index to use (default: records)
+  [VERIFY_CERTS]: verify certs of Elasticsearch instance (default: True)
+  [HASH_RECORD]: make record unique by hashing record [slow] (default: False)
 """
 
 log = logging.getLogger(__name__)
@@ -36,12 +37,14 @@ class ElasticWriter(AbstractWriter):
         index: str = "records",
         verify_certs: Union[str, bool] = True,
         http_compress: Union[str, bool] = True,
+        hash_record: Union[str, bool] = False,
         **kwargs,
     ) -> None:
         self.index = index
         self.uri = uri
         verify_certs = str(verify_certs).lower() in ("1", "true")
         http_compress = str(http_compress).lower() in ("1", "true")
+        self.hash_record = str(hash_record).lower() in ("1", "true")
         self.es = elasticsearch.Elasticsearch(uri, verify_certs=verify_certs, http_compress=http_compress)
         self.json_packer = JsonRecordPacker()
         self.queue: queue.Queue[Union[Record, StopIteration]] = queue.Queue()
@@ -50,7 +53,7 @@ class ElasticWriter(AbstractWriter):
         self.thread.start()
 
         if not verify_certs:
-            # disable InsecureRequestWarning of urllib3, caused by the verify_certs flag
+            # Disable InsecureRequestWarning of urllib3, caused by the verify_certs flag.
             import urllib3
             urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
@@ -75,6 +78,14 @@ class ElasticWriter(AbstractWriter):
             "_index": index,
             "_source": self.json_packer.pack(rdict),
         }
+
+        # Check if hash_record is set and hash record to md5 if flag is set.
+        # Note: the _record_metadata contains timestamp field of target-query output.
+        # An option to change this is to call the `self.json_packer.pack` again on the
+        # `record._asdict()` row.
+        if self.hash_record:
+            document["_id"] = hashlib.md5(document["_source"].encode()).hexdigest()
+
         return document
 
     def document_stream(self) -> Iterator[dict]:
@@ -128,7 +139,7 @@ class ElasticReader(AbstractReader):
         self.es = elasticsearch.Elasticsearch(uri, verify_certs=verify_certs, http_compress=http_compress)
 
         if not verify_certs:
-            # disable InsecureRequestWarning of urllib3, caused by the verify_certs flag
+            # Disable InsecureRequestWarning of urllib3, caused by the verify_certs flag.
             import urllib3
             urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
--- a/flow/record/adapter/elastic.py
+++ b/flow/record/adapter/elastic.py
@@ -20,22 +20,39 @@ Read usage: rdump elastic+[PROTOCOL]://[IP]:[PORT]?index=[INDEX]
 [IP]:[PORT]: ip and port to elastic host
 [INDEX]: index to write to or read from
 [PROTOCOL]: http or https. Defaults to https when "+[PROTOCOL]" is omitted
+
+Optional arguments:
+  [INDEX]: name of the index to use
+  [VERIFY_CERTS]: verify certs of Elasticsearch instance
 """
 
 log = logging.getLogger(__name__)
 
 
 class ElasticWriter(AbstractWriter):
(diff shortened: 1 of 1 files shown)


0xbart:
@DissectBot agree

0xbart:
I also have another option to introduce, called `hash_record`. The data can be ingested multiple times, such as caused by an HTTP error. An option to prevent this is to make the `_id` unique. One of the ways to do this is by using the `hashlib.md5` function on the `document["_source"]` variable. 

Passing this option is a bit slow, but I think useful to introduce. Interesting to hear other opinions.

Example code:
```
    def record_to_document(self, record: Record, index: str) -> dict:
        """Convert a record to a Elasticsearch compatible document dictionary"""
        rdict = record._asdict()

        (...)

        # Check if hash_record is set and hash record to md5 if flag is set
        if self.hash_record:
            document["_id"] = hashlib.md5(
                document["_source"].encode()
                # self.json_packer.pack(record._asdict()).encode().  
                # ^ This is also an option if `target-query` is used again
            ).hexdigest()

        return document
```

Oneliner to test: `rdump <records> -w 'elastic+https://@127.0.0.1:9200?verify_certs=0&hash_record=1`

flow/record/adapter/elastic.py
```diff
@@ -87,25 +104,33 @@ def flush(self) -> None:
         pass
 
     def close(self) -> None:
-        self.queue.put(StopIteration)
-        self.event.wait()
-        self.es.close()
+        if hasattr(self, "es"):
```
yunzheng, line 107: If `elasticsearch.Elasticsearch()` can indeed return `None`, we should just raise an exception in`__init__()`. Do you have an example uri that would make `self.es = None`?

flow/record/adapter/elastic.py
```diff
@@ -121,4 +146,5 @@ def __iter__(self) -> Iterator[Record]:
                 yield obj
 
     def close(self) -> None:
-        self.es.close()
+        if hasattr(self, "es"):
```
yunzheng, line 149: these hasattr guards would be unnecessary if we check `self.es` in `__init__()`

flow/record/adapter/elastic.py
```diff
@@ -20,23 +20,40 @@
 [IP]:[PORT]: ip and port to elastic host
 [INDEX]: index to write to or read from
 [PROTOCOL]: http or https. Defaults to https when "+[PROTOCOL]" is omitted
+
+Optional arguments:
+  [INDEX]: name of the index to use
+  [VERIFY_CERTS]: verify certs of Elasticsearch instance
```
yunzheng, line 26: I like the Optional arguments section, which makes it a bit more readable. We can also adopt this in other adapters.

However, now `[INDEX]` is listed twice. I made a code suggestion below, which als adds defaults:
```suggestion
[PROTOCOL]: http or https. Defaults to https when "+[PROTOCOL]" is omitted

Optional arguments:
  [INDEX]: name of the index to use (default: records)
  [VERIFY_CERTS]: verify certs of Elasticsearch instance (default: True)
```

yunzheng (changes requested):

yunzheng:
> I also have another option to introduce, called `hash_record`. The data can be ingested multiple times, such as caused by an HTTP error. An option to prevent this is to make the `_id` unique. One of the ways to do this is by using the `hashlib.md5` function on the `document["_source"]` variable.
> 
> Passing this option is a bit slow, but I think useful to introduce. Interesting to hear other opinions.
> 

It's definitely a nice idea to have a unique (deterministic) identifier for a record, and could also be useful for other ingestors to avoid duplication. Do note that the `_generated` timestamp field will make the record non deterministic (unless you replay it from archived records), but if this is excluded from the hashing you would have a deterministic identifier.

It would be better if this deterministic id generation would happen in the core of flow.record, but i'm not sure yet what the best way would be to achieve this. 

So for now, I think it's fine to have this feature in the elastic adapter. Feel free to add to this PR or create a new one.

flow/record/adapter/elastic.py
0xbart: The `elasticsearch.Elasticsearch()` does not return `None`, but raise an ValueError with invalid scheme. The deconstructor function of the class then calls the `close()` function, which contains function calls on undefined variables. See the comment for the stack strace.

0xbart: This is also why variables are defined in other adapters as well. Such as the `splunk.py` where the `self.sock` is set. Its also possible to use a class variable inside this file, but I prefer the way of `hasattr` above the class variables.

flow/record/adapter/elastic.py
0xbart: Good idea. Will update.

0xbart:
@yunzheng Thanks for your review! I've updated the PR with the `hash_record` flag. 

The reason why I've chosen for the `hasattr()` is because an Exception is raised during the Exception. See example of the stack trace: 

```
Traceback (most recent call last):
  File "/home/user/.virtualenvs/dissect/src/flow-record/flow/record/tools/rdump.py", line 239, in <module>
    sys.exit(main())
  File "/home/user/.local/share/virtualenvs/dissect/src/flow-record/flow/record/utils.py", line 57, in wrapper
    return func(*args, **kwargs)
  File "/home/user/.virtualenvs/dissect/src/flow-record/flow/record/tools/rdump.py", line 209, in main
    with RecordWriter(uri) as record_writer:
  File "/home/user/.local/share/virtualenvs/dissect/src/flow-record/flow/record/base.py", line 864, in RecordWriter
    return RecordAdapter(url=url, out=True, clobber=clobber, **kwargs)
  File "/home/user/.local/share/virtualenvs/dissect/src/flow-record/flow/record/base.py", line 851, in RecordAdapter
    return cls(cls_url, **arg_dict)
  File "/home/user/.local/share/virtualenvs/dissect/src/flow-record/flow/record/adapter/elastic.py", line 49, in __init__
    self.es = elasticsearch.Elasticsearch(uri, verify_certs=verify_certs, http_compress=http_compress)
  File "/home/user/.local/share/virtualenvs/dissect/lib/python3.10/site-packages/elasticsearch/_sync/client/__init__.py", line 333, in __init__
    node_configs = client_node_configs(
  File "/home/user/.local/share/virtualenvs/dissect/lib/python3.10/site-packages/elasticsearch/_sync/client/utils.py", line 108, in client_node_configs
    node_configs = hosts_to_node_configs(hosts)
  File "/home/user/.local/share/virtualenvs/dissect/lib/python3.10/site-packages/elasticsearch/_sync/client/utils.py", line 146, in hosts_to_node_configs
    return hosts_to_node_configs([hosts])
  File "/home/user/.local/share/virtualenvs/dissect/lib/python3.10/site-packages/elasticsearch/_sync/client/utils.py", line 154, in hosts_to_node_configs
    node_configs.append(url_to_node_config(host))
  File "/home/user/.local/share/virtualenvs/dissect/lib/python3.10/site-packages/elastic_transport/client_utils.py", line 216, in url_to_node_config
    raise ValueError(
ValueError: URL must include a 'scheme', 'host', and 'port' component (ie 'https://localhost:9200')

Exception ignored in: <function AbstractWriter.__del__ at 0x7fd97ad0d750>
Traceback (most recent call last):
  File "/home/user/.local/share/virtualenvs/dissect/src/flow-record/flow/record/adapter/__init__.py", line 39, in __del__
    self.close()
  File "/home/user/.local/share/virtualenvs/dissect/src/flow-record/flow/record/adapter/elastic.py", line 119, in close
    self.queue.put(StopIteration)
AttributeError: 'ElasticWriter' object has no attribute 'queue'
```

flow/record/adapter/elastic.py
yunzheng: gotcha!

flow/record/adapter/elastic.py
```diff
@@ -58,6 +78,14 @@ def record_to_document(self, record: Record, index: str) -> dict:
             "_index": index,
             "_source": self.json_packer.pack(rdict),
         }
+
+        # Check if hash_record is set and hash record to md5 if flag is set.
+        # Note: the _record_metadata contains timestamp field of target-query output.
+        # An option to change this is to call the `self.json_packer.pack` again on the
+        # `record._asdict()` row.
```
yunzheng, line 85: What do you mean with this timestamp comment?

yunzheng:
There are some linting errors, you can fix them with `tox -e fix`

0xbart:
> There are some linting errors, you can fix them with `tox -e fix`

Done.

flow/record/adapter/elastic.py
0xbart: Nevermind. Removed for now. Maybe later a code change for dealing a different way with the `_generated` timestamp field.

yunzheng:
https://github.com/fox-it/flow.record/blob/5cb258d0f9f963d5a44dfa9f85feaaa0d00c55a5/flow/record/adapter/elastic.py#L72-L76

Can you add the following between line 75 and 76:

```python
        # remove _generated field from metadata to ensure determinstic documents
        if self.hash_record:
            rdict_meta.pop("generated", None)
```

This will ensure that the record document is deterministic by removing the `_generated`  field when `hash_record` is set.

yunzheng (approved):
LGTM, thanks for the contribution!

PR #93, opened 2023-10-26
Fix str() and repr() in selectors

Schamper:
This allows selectors such as `"my_string" in str(r)` for work, for when you want to "grep" a selector.

(DIS-2565)


yunzheng (changes requested):
Can you also add `__repr__` and also add one negative test?

yunzheng (approved):
LGTM

PR #96, opened 2023-11-16
Refactor and improve CSV adapter

yunzheng:
This change allows the CSV adapter to:
 - properly read reserved field names (eg: _generated, _source, etc)
 - use `clean_fieldname` that otherwise are invalid names in flow.record
 - deduce format of csv file automatically by using `csv.Sniffer`

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -982,9 +982,9 @@ def normalize_fieldname(field_name: str) -> str:
         >>> normalize_fieldname("my-variable-name-with-dashes")
         'my_variable_name_with_dashes'
         >>> normalize_fieldname("_my_name_starting_with_underscore")
-        'n__my_name_starting_with_underscore'
+        'x__my_name_starting_with_underscore'
         >>> normalize_fieldname("1337")
-        'n_1337'
+        'x_1337'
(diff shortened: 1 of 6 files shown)


flow/record/base.py
```diff
@@ -971,6 +971,36 @@ def extend_record(
...
+        'my_name_with_spaces'
+        >>> normalize_fieldname("my name (with) parentheses")
+        'my_name__with__parentheses'
+        >>> normalize_fieldname("_generated")
+        '_generated'
+    """
+
+    if field_name not in RESERVED_FIELDS:
+        field_name = re.sub(r"[- ()]", "_", field_name)
+        # prepend `n_` if field_name is empty or starts with underscore or digit
```
Schamper, line 998: Why `n_`?

Schamper (dismissed):

flow/record/base.py
yunzheng: This was the original prefix if the field started with a number, i've kept it since. If you know a better one, let me know :)

Schamper: `x_`?

yunzheng: done!

Schamper (approved):

PR #97, opened 2023-11-21
Add DuckDB adapter

yunzheng:
This adds DuckDB reader and writer support. Because DuckDB is mostly
compatible with the SQLite API, we just subclass from the existing
SQLite adapter with minimal changes.

Changes done to the SQLite adapter and tests are:

 * backtick quoting does not work in DuckDB so we use double quotes now
 * DuckDB has strict typing so some tests are not applicable and skipped
 * `executescript()` does not exist in DuckDB so we avoid using it
 * Switched SQLite to `isolation_level=None` for manual transactions


flow/record/adapter/sqlite.py
```diff
@@ -58,12 +56,12 @@ def create_descriptor_table(con: sqlite3.Connection, descriptor: RecordDescripto
...
         column_type = FIELD_MAP.get(fieldset.typename, "TEXT")
-        column_defs.append(f"   `{column_name}` {column_type}")
+        column_defs.append(f'   "{column_name}" {column_type}')
     sql_columns = ",\n".join(column_defs)
 
     # Create the descriptor table
-    sql = f"CREATE TABLE IF NOT EXISTS `{table_name}` (\n{sql_columns}\n)"
-    logger.debug(sql)
+    sql = f'CREATE TABLE IF NOT EXISTS "{table_name}" (\n{sql_columns}\n)'
+    logging.debug(sql)
```
Schamper, line 64: Isn't it recommended to use a module specific logger?

flow/record/adapter/sqlite.py
```diff
@@ -141,6 +141,7 @@ def read_table(self, table_name: str) -> Iterator[Record]:
         # flow.record is quite strict with what is allowed in fieldnames or decriptor name.
         # While SQLite is less strict, we need to sanitize the names to make them compatible.
         table_name_org = table_name
+        table_name_org = table_name_org.replace('"', '""')
```
Schamper, line 144:
```suggestion
        table_name_org = table_name.replace('"', '""')
```
Also what effect does this have exactly? An insert statement with a table name of `a"b` will look like `INSERT INTO "a""b"`? How does that work? Is this some escaping behaviour of SQLite/DuckDB?

flow/record/adapter/sqlite.py
```diff
@@ -193,12 +194,15 @@ def __iter__(self) -> Iterator[Record]:
...
+    """SQLite writer."""
+
+    def __init__(self, path: str, *, batch_size: str | int = 1000, **kwargs):
         self.descriptors_seen = set()
         self.con = None
-        self.con = sqlite3.connect(path)
+        self.con = sqlite3.connect(path, isolation_level=None)
         self.count = 0
         self.batch_size = int(batch_size)
+        self.con.execute("BEGIN")
```
Schamper, line 205:
```suggestion
        self.tx_cycle()
```
?

flow/record/adapter/duckdb.py
```diff
@@ -0,0 +1,48 @@
...
+class DuckdbWriter(SqliteWriter):
+    """DuckDB writer, subclasses from SQLite writer."""
+
+    def __init__(self, path: str, *, batch_size: str | int = 1000, **kwargs):
+        self.descriptors_seen = set()
+        self.con = None
+        self.con = duckdb.connect(path)
+        self.count = 0
+        self.batch_size = int(batch_size)
+        self.con.begin()
```
Schamper, line 44: What's the difference between this and `self.con.execute("BEGIN")`?

flow/record/adapter/duckdb.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def tx_cycle(self) -> None:
+        self.con.execute("COMMIT")
```
Schamper, line 47: Can't check if you're in a transaction with DuckDB?

Schamper (changes requested):

flow/record/adapter/sqlite.py
yunzheng: I changed it to global due to subclassing, but I think i'll just add a logger to the class.

flow/record/adapter/duckdb.py
yunzheng: effectively it's the same. I will just change it to `tx_cycle()` as well to make it consistent.

flow/record/adapter/duckdb.py
yunzheng: Sadly no.

flow/record/adapter/sqlite.py
yunzheng: This is indeed escaping behaviour that works both on SQLite and DuckDB:

```shell
> duckdb
v0.8.1 6536a77232
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D create table 'a"b' (foo text);
D .schema
CREATE TABLE IF NOT EXISTS "a""b"(foo VARCHAR);;
D insert into "a""b" values ('bar');
D select * from "a""b";
┌─────────┐
│   foo   │
│ varchar │
├─────────┤
│ bar     │
└─────────┘
D
```

```shell
> sqlite3
SQLite version 3.28.0 2019-04-15 14:49:49
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> create table 'a"b' (foo text);
sqlite> .schema
CREATE TABLE IF NOT EXISTS 'a"b' (foo text);
sqlite> insert into "a""b" values ('bar');
sqlite> select * from "a""b";
bar
sqlite>
```

flow/record/adapter/duckdb.py
yunzheng: actually I can't, as sqlite3 has no `begin()` method.

flow/record/adapter/duckdb.py
Schamper, line 1: Any idea why GitHub says there are no tests for this file?

tests/test_sqlite_duckdb_adapter.py
```diff
@@ -308,41 +340,49 @@ def test_prepare_insert_sql():
...
-    with RecordWriter(f"sqlite://{db_path}?batch_size={batch_size}") as writer:
+    with RecordWriter(f"{db.scheme}://{db_path}?batch_size={batch_size}") as writer:
         # write a single record, should not be flushed yet if batch_size > 1
         writer.write(next(records))
 
         # test count of records in table (no flush yet if batch_size > 1)
-        with sqlite3.connect(db_path) as con:
-            x = con.execute("select count(*) from `test/record`")
+        with db.connector.connect(str(db_path)) as con:
+            x = con.execute('select count(*) from "test/record"')
```
Schamper, line 360:
```suggestion
            x = con.execute('SELECT COUNT(*) FROM "test/record"')
```
Might be nice to be consistent.

tests/test_sqlite_duckdb_adapter.py
```diff
@@ -308,41 +340,49 @@ def test_prepare_insert_sql():
...
 
         # write at least batch_size records, should be flushed due to batch_size
         for i in range(batch_size):
             writer.write(next(records))
 
         # test count of records in table after flush
-        with sqlite3.connect(db_path) as con:
-            x = con.execute("select count(*) from `test/record`")
+        with db.connector.connect(str(db_path)) as con:
+            x = con.execute('select count(*) from "test/record"')
```
Schamper, line 369:
```suggestion
            x = con.execute('SELECT COUNT(*) FROM "test/record"')
```

Schamper (changes requested):

flow/record/adapter/sqlite.py
```diff
@@ -123,7 +126,11 @@ def db_insert_record(con: sqlite3.Connection, record: Record) -> None:
 
 
 class SqliteReader(AbstractReader):
-    def __init__(self, path: str, batch_size: str | int = 1000, selector: Selector | str | None = None, **kwargs):
+    """SQLite reader."""
+
+    logger = logger
```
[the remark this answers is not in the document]
yunzheng: It’s mainly if the logging config uses `%(name)s` the correct module name is shown.

flow/record/adapter/duckdb.py
yunzheng: No, i’ll investigate

yunzheng: I think it's probably due to dynamic imports of the adapter, will try to rewrite the tests and see if that helps

yunzheng: It was actually a tox version issue. Latest tox version 4.11.4 seems to properly parse the pyproject.toml via the `.[test]` extras.

Schamper: @pyrco any objection to updating the minimal tox version?

yunzheng: While the latest tox version (4.11.4) does seem to parse the `.[test]` extras better, it doesn't seem to work with conditionals, for example it still installed the following on PyPy using the following `pyproject.toml`:

```
[project.optional-dependencies]
duckdb = [
    "duckdb",
    "pytz", # duckdb requires pytz for timezone support
]
test = [
    "flow.record[duckdb]; platform_python_implementation != 'PyPy'",
]
```

However, with the older tox (version 4.2.4) it would not install duckdb at all.

I've now opted to just not self reference extras in `.[test]` in a newer commit.

yunzheng:
DuckDB is excluded for now on Python 3.12 as there are no wheels yet. When they are it can be readded in the tests.

yunzheng:
It looks like codecov uses Python 3.12 for the coverage tests and on Python 3.12 I skip the duckdb tests, not sure where I can check that.

yunzheng:
@Schamper New duckdb release seems to have fixed the codecov tests.

Schamper (approved):

PR #101, opened 2023-12-19
Use reprlib to limit the warning message

yunzheng:
The warning message could get very long (single line) due to extra data in msgpack parsing. For example due to corrupt file, or incorrect use.

By using reprlib we limit the length of the warning message.

### DIFF
--- a/flow/record/stream.py
+++ b/flow/record/stream.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 import datetime
 import logging
 import os
+import reprlib
 import struct
 import sys
 from collections import ChainMap
@@ -16,6 +17,9 @@ from .base import RecordDescriptor, RecordReader
 from .packer import RecordPacker
 
 log = logging.getLogger(__package__)
+
+aRepr = reprlib.Repr()
+aRepr.maxother = 100
 
 
 def RecordOutput(fp):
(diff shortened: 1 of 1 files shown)


flow/record/stream.py
```diff
@@ -17,6 +18,9 @@
 
 log = logging.getLogger(__package__)
 
+aRepr = reprlib.Repr()
+aRepr.maxother = 255
```
Schamper, line 22:
```suggestion
aRepr = reprlib.Repr(maxother=255)
```
?

Schamper (changes requested):

flow/record/stream.py
yunzheng: Sadly Repr() doesn't have a initializer for this:

```python
Init signature: reprlib.Repr()
Docstring:      <no docstring>
Source:
class Repr:

    def __init__(self):
        self.maxlevel = 6
        self.maxtuple = 6
        self.maxlist = 6
        self.maxarray = 5
        self.maxdict = 4
        self.maxset = 6
        self.maxfrozenset = 6
        self.maxdeque = 6
        self.maxstring = 30
        self.maxlong = 40
        self.maxother = 30
```

Schamper: Oh I missed this line in the documentation:
> Changed in version 3.12: Allow attributes to be set via keyword arguments.

Schamper (approved):

PR #107, opened 2024-02-14
Make records hashable

JSCU-CNI:
This PR implements the `__hash__` functions for the `Record` class, enabling grouping records in sets. This enables set operations on collections of records, e.g:

```python
records_a = set(target_one.users())
records_b = set(target_two.users())

diff = records_a.symmetric_difference(records_b)
```

For some record types though, there will be fields that you might want to ignore, such as `_generated` and possibly `_source` when comparing record output from two different targets. Preferably you want the fields to ignore to be configurable at runtime. Therefore, we opted to add an environment variable `FLOW_RECORD_IGNORE` which can contain a list of comma-separated field names to ignore. 

We implemented a check where hashed records are considered 'immutable' and a `ValueError` is raised if a property is changed after `__hash__` has been called on a `Record` instance. This may have some performance impact though. Feel free to suggest alternative approaches.

### DIFF
--- a/flow/record/__init__.py
+++ b/flow/record/__init__.py
@@ -21,8 +21,8 @@ from flow.record.base import (
     open_path,
     open_path_or_stream,
     open_stream,
+    set_ignored_fields_for_comparison,
     stream,
-    update_ignored_fields_for_comparison,
 )
 from flow.record.jsonpacker import JsonRecordPacker
 from flow.record.stream import (
@@ -57,8 +57,8 @@ __all__ = [
     "open_path_or_stream",
     "open_path",
     "open_stream",
+    "set_ignored_fields_for_comparison",
     "stream",
-    "update_ignored_fields_for_comparison",
     "dynamic_fieldtype",
     "DynamicDescriptor",
     "PathTemplateWriter",
--- a/flow/record/__init__.py
+++ b/flow/record/__init__.py
@@ -2,6 +2,7 @@ import gzip
 import os
 
 from flow.record.base import (
+    IGNORE_FIELDS_FOR_COMPARISON,
     RECORD_VERSION,
     RECORDSTREAM_MAGIC,
     DynamicDescriptor,
@@ -21,6 +22,7 @@ from flow.record.base import (
     open_path_or_stream,
     open_stream,
     stream,
+    update_ignored_fields_for_comparison,
 )
 from flow.record.jsonpacker import JsonRecordPacker
 from flow.record.stream import (
@@ -35,6 +37,7 @@ from flow.record.stream import (
 )
 
 __all__ = [
+    "IGNORE_FIELDS_FOR_COMPARISON",
     "RECORD_VERSION",
     "RECORDSTREAM_MAGIC",
     "FieldType",
@@ -55,6 +58,7 @@ __all__ = [
     "open_path",
     "open_stream",
     "stream",
+    "update_ignored_fields_for_comparison",
     "dynamic_fieldtype",
     "DynamicDescriptor",
     "PathTemplateWriter",
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -15,7 +15,17 @@ import warnings
 from datetime import datetime, timezone
 from itertools import zip_longest
 from pathlib import Path
-from typing import IO, Any, BinaryIO, Iterator, Mapping, Optional, Sequence, Union
+from typing import (
+    IO,
+    Any,
+    BinaryIO,
+    Iterable,
+    Iterator,
+    Mapping,
+    Optional,
+    Sequence,
+    Union,
+)
 from urllib.parse import parse_qsl, urlparse
 
 from flow.record.adapter import AbstractReader, AbstractWriter
@@ -97,15 +107,15 @@ class {name}(Record):
 
 
 if env_excluded_fields := os.environ.get("FLOW_RECORD_IGNORE"):
-    IGNORE_FIELDS_FOR_COMPARISON = env_excluded_fields.split(",")
+    IGNORE_FIELDS_FOR_COMPARISON = set(env_excluded_fields.split(","))
 else:
-    IGNORE_FIELDS_FOR_COMPARISON = []
+    IGNORE_FIELDS_FOR_COMPARISON = set()
 
 
-def update_ignored_fields_for_comparison(ignored_fields: list[str]) -> None:
+def set_ignored_fields_for_comparison(ignored_fields: Iterable[str]) -> None:
     """Can be used to update the IGNORE_FIELDS_FOR_COMPARISON from outside the flow.record package scope"""
     global IGNORE_FIELDS_FOR_COMPARISON
-    IGNORE_FIELDS_FOR_COMPARISON = ignored_fields
+    IGNORE_FIELDS_FOR_COMPARISON = set(ignored_fields)
 
 
 class FieldType:
@@ -179,7 +189,7 @@ class Record:
         return result
 
     def __hash__(self) -> int:
-        return self._pack(excluded_fields=IGNORE_FIELDS_FOR_COMPARISON).__hash__()
+        return hash(self._pack(excluded_fields=IGNORE_FIELDS_FOR_COMPARISON))
 
     def __repr__(self):
         return "<{} {}>".format(
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -95,6 +95,8 @@ class {name}(Record):
 {unpack_code}
 """
 
+IGNORE_FIELDS_FOR_COMPARISON_VARIABLE = "FLOW_RECORD_IGNORE"
+
 
 class FieldType:
     def _typename(self):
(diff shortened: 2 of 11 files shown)


flow/record/base.py
```diff
@@ -111,36 +113,52 @@ def _unpack(cls, data):
...
+                continue
+
             # Skip version field if requested (only for compatibility reasons)
             if unversioned and k == "_version" and v == 1:
                 continue
-            else:
-                values.append(v)
+
+            if k == "_mutable":
+                continue
```
yunzheng, line 147: See main review comment about `_mutable` flag/field.

flow/record/base.py
```diff
@@ -111,36 +113,52 @@ def _unpack(cls, data):
...
+
+            values.append(v)
 
         return self._desc.identifier, tuple(values)
 
     def _packdict(self):
         return dict(
             (k, v._pack() if isinstance(v, FieldType) else v)
-            for k, v in ((k, getattr(self, k)) for k in self.__slots__)
+            for k, v in ((k, getattr(self, k)) for k in self.__slots__ if k != "_mutable")
```
yunzheng, line 156: See main review comment about `_mutable` flag/field.

flow/record/base.py
```diff
@@ -111,36 +113,52 @@ def _unpack(cls, data):
...
         )
 
     def _asdict(self, fields=None, exclude=None):
         exclude = exclude or []
+        exclude.append("_mutable")
```
yunzheng, line 161: See main review comment about `_mutable` flag/field.

flow/record/base.py
```diff
@@ -244,17 +273,19 @@ def __getattr__(self, attr):
...
+            tuple(record._pack(excluded_fields) for record in self.records),
         )
 
     def _replace(self, **kwds):
         new_records = []
         for record in self.records:
             new_records.append(
-                record.__class__(*map(kwds.pop, record.__slots__, (getattr(self, k) for k in record.__slots__)))
+                record.__class__(
+                    *map(kwds.pop, record.__slots__, (getattr(self, k) for k in record.__slots__ if k != "_mutable"))
```
yunzheng, line 287: See main review comment about `_mutable` flag/field.

flow/record/base.py
```diff
@@ -355,13 +386,16 @@ def _generate_record_class(name: str, fields: tuple[tuple[str, str]]) -> type:
 
     name = name.replace("/", "_")
     args = ""
-    init_code = ""
+    init_code = "\t\t__self._mutable = True\n"
```
yunzheng, line 389: See main review comment about `_mutable` flag/field.

flow/record/base.py
```diff
@@ -396,7 +430,7 @@ def _generate_record_class(name: str, fields: tuple[tuple[str, str]]) -> type:
     code = RECORD_CLASS_TEMPLATE.format(
         name=name,
         args=args,
-        slots_tuple=tuple(all_fields.keys()),
+        slots_tuple=(tuple(all_fields.keys()) + ("_mutable",)),
```
yunzheng, line 433: See main review comment about `_mutable` flag/field.

flow/record/base.py
```diff
@@ -542,7 +576,7 @@ def init_from_dict(self, rdict: dict[str, Any], raise_unknown=False) -> Record:
         """
 
         if not raise_unknown:
-            rdict = {k: v for k, v in rdict.items() if k in self.recordType.__slots__}
+            rdict = {k: v for k, v in rdict.items() if k in self.recordType.__slots__ and k != "_mutable"}
```
yunzheng, line 579: See main review comment about `_mutable` flag/field.

tests/test_record.py
```diff
@@ -792,3 +798,28 @@ def test_normalize_fieldname():
...
+    )
+
+    same_same = TestRecord(firstname="James", lastname="Bond")
+    but_different = TestRecord(firstname="Ethan", lastname="Hunt")
+    but_still_same = TestRecord(firstname="Andrew", lastname="Bond")
+
+    records = [same_same, but_different, but_still_same]
+
+    assert same_same != but_still_same
+    with _patch_dict(os.environ, {IGNORE_FIELDS_FOR_COMPARISON_VARIABLE: "_generated,firstname"}):
```
yunzheng, line 819: would also be nice to add an extra test by setting the global variable

yunzheng (changes requested):
Regarding the `_mutable` flag/field, I think it's best if we just treat it as how dataclasses works when using `unsafe_hash=True`, hashable but still mutable:

```python
In [1]: from dataclasses import dataclass
   ...:
   ...: @dataclass(unsafe_hash=True)
   ...: class MyRecord:
   ...:     foo: str
   ...:     bar: int
   ...:

In [2]: a = MyRecord("a", 1337)

In [3]: b = MyRecord("b", 1337)

In [4]: set([a, b])
Out[4]: {MyRecord(foo='a', bar=1337), MyRecord(foo='b', bar=1337)}

In [5]: b.foo = "a"

In [6]: set([a, b])
Out[6]: {MyRecord(foo='a', bar=1337)}
```

I think the `_mutable` flag/field introduces unnecessary hand holding and extra code complexity. So it's best if we just not implement this.

Ideally we want to make flow.record compatible with dataclasses which could solve most of these features more nicely. It's something we want to explore in the future.

flow/record/base.py
```diff
@@ -160,6 +178,9 @@ def _replace(self, **kwds):
             raise ValueError("Got unexpected field names: {kwds!r}".format(kwds=list(kwds)))
         return result
 
+    def __hash__(self) -> int:
+        return self._pack(excluded_fields=IGNORE_FIELDS_FOR_COMPARISON).__hash__()
```
yunzheng, line 182:
```suggestion
        return hash(self._pack(excluded_fields=IGNORE_FIELDS_FOR_COMPARISON))
```

flow/record/base.py
```diff
@@ -96,6 +96,18 @@ def _unpack(__cls, {args}):
...
+if env_excluded_fields := os.environ.get("FLOW_RECORD_IGNORE"):
+    IGNORE_FIELDS_FOR_COMPARISON = env_excluded_fields.split(",")
+else:
+    IGNORE_FIELDS_FOR_COMPARISON = []
+
+
+def update_ignored_fields_for_comparison(ignored_fields: list[str]) -> None:
+    """Can be used to update the IGNORE_FIELDS_FOR_COMPARISON from outside the flow.record package scope"""
+    global IGNORE_FIELDS_FOR_COMPARISON
+    IGNORE_FIELDS_FOR_COMPARISON = ignored_fields
```
yunzheng, line 108: I think `set_ignored_fields_for_comparison()` is a bit more clearer than `update_ignored_fields_for_comparison()`. The latter would make me think that it does something with the previous value.

Also using `set()` for efficiency and filter out possible duplicate field names.

```suggestion
def set_ignored_fields_for_comparison(ignored_fields: list[str]) -> None:
    """Can be used to update the IGNORE_FIELDS_FOR_COMPARISON from outside the flow.record package scope"""
    global IGNORE_FIELDS_FOR_COMPARISON
    IGNORE_FIELDS_FOR_COMPARISON = set(ignored_fields)
```

flow/record/base.py
```diff
@@ -96,6 +96,18 @@ def _unpack(__cls, {args}):
 """
 
 
+if env_excluded_fields := os.environ.get("FLOW_RECORD_IGNORE"):
+    IGNORE_FIELDS_FOR_COMPARISON = env_excluded_fields.split(",")
+else:
+    IGNORE_FIELDS_FOR_COMPARISON = []
```
yunzheng, line 102:
```suggestion
if env_excluded_fields := os.environ.get("FLOW_RECORD_IGNORE"):
    IGNORE_FIELDS_FOR_COMPARISON = set(env_excluded_fields.split(","))
else:
    IGNORE_FIELDS_FOR_COMPARISON = set()
```

flow/record/__init__.py
```diff
@@ -21,6 +22,7 @@
     open_path_or_stream,
     open_stream,
     stream,
+    update_ignored_fields_for_comparison,
```
yunzheng, line 25:
```suggestion
    set_ignored_fields_for_comparison,
```

flow/record/__init__.py
```diff
@@ -55,6 +58,7 @@
     "open_path",
     "open_stream",
     "stream",
+    "update_ignored_fields_for_comparison",
```
yunzheng, line 61:
```suggestion
    "set_ignored_fields_for_comparison",
```

tests/test_record.py
```diff
@@ -792,3 +799,60 @@ def test_normalize_fieldname():
...
+
+    same_same = TestRecord(firstname="James", lastname="Bond")
+    but_different = TestRecord(firstname="Ethan", lastname="Hunt")
+    but_still_same = TestRecord(firstname="Andrew", lastname="Bond")
+
+    records = [same_same, but_different, but_still_same]
+
+    assert same_same != but_still_same
+
+    update_ignored_fields_for_comparison(["_generated", "firstname"])
```
yunzheng, line 821:
```suggestion
    set_ignored_fields_for_comparison(["_generated", "firstname"])
```

yunzheng (changes requested):

tests/test_record.py
```diff
@@ -792,3 +799,60 @@ def test_normalize_fieldname():
...
+        # Force a re-import of flow.record so the global variable gets re-initialized based on the environment variable
+        keys = [key for key in sys.modules if key == "flow" or "flow." in key]
+        for key in keys:
+            del sys.modules[key]
+
+        importlib.import_module("flow.record")
+
+        from flow.record import IGNORE_FIELDS_FOR_COMPARISON, RecordDescriptor
+
+        assert IGNORE_FIELDS_FOR_COMPARISON == ["_generated", "lastname"]
```
yunzheng, line 840:
```suggestion
        assert IGNORE_FIELDS_FOR_COMPARISON == {"_generated", "lastname"}
```

yunzheng (approved):
Thanks for the nice feature and implementation, LGTM

PR #108, opened 2024-02-15
Make records sortable

JSCU-CNI:
This PR adds sorting functionality to the flow.record `Record` base class.

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -119,6 +119,9 @@ class Record:
             return False
         return self._pack() == other._pack()
 
+    def __lt__(self, other):
+        return self._generated < other._generated
+
     def _pack(self, unversioned=False):
         values = []
         for k in self.__slots__:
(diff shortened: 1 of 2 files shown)


yunzheng:
Sorting on `_generated` is universal across multiple different RecordDescriptors as this field should always exists, however i'm not sure if this is directly what people expect when you sort a collection of records and could lead to unexpected behaviour.

Looking at dataclasses, it's not sortable by default either (unless you specify `order=True`, and then it would sort on the tuples), but then again it's not sortable across different dataclasses. 

@Schamper do you have any opinions on this?

Schamper:
Maybe another question first, what is the primary use-case for this feature? If it was the unit test of https://github.com/fox-it/dissect.target/pull/507, then this addition wouldn't have resolved that. Is there another reason why this should be added?

If we do add this, I think the dataclass approach sounds sane (only same type, on tuples).

PR #109, opened 2024-02-20
Move lru_cache definitions to __init__

Miauwkeru:
Using the lru_cache decorators on class methods, the ones that have a reference to `self`,
will also cache self. So we move it to the __init__ of the class
    
(DIS-2913)

### DIFF
--- a/flow/record/stream.py
+++ b/flow/record/stream.py
@@ -256,7 +256,7 @@ class RecordFieldRewriter:
         self.exclude = exclude or []
         self.expression = compile(expression, "<string>", "exec") if expression else None
 
-        self.record_descriptor_for_fields = lru_cache(maxsize=256)(self.record_descriptor_for_fields)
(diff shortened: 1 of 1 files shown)


flow/record/stream.py
```diff
@@ -256,7 +256,8 @@ def __init__(self, fields=None, exclude=None, expression=None):
         self.exclude = exclude or []
         self.expression = compile(expression, "<string>", "exec") if expression else None
 
-    @lru_cache(maxsize=256)
+        self.record_descriptor_for_fields = lru_cache(maxsize=256)(self.record_descriptor_for_fields)
```
Schamper, line 259:
```suggestion
        self.record_descriptor_for_fields = lru_cache(256)(self.record_descriptor_for_fields)
```

Schamper (changes requested):

Schamper (approved):

PR #116, opened 2024-04-10
Fix hashing of records with list fieldtypes

JSCU-CNI:
Currently flow.record breaks when `__hash__` is called on a record that has a list fieldtype. As lists are mutable, they have to be converted into tuples first so they can be hashed.

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -215,7 +215,7 @@ class Record:
             for list_value in value:
                 if isinstance(list_value, dict):
                     # List values that are dicts must be converted to tuples
-                    dict_as_tuple = ((k, v) for k, v in list_value.items())
+                    dict_as_tuple = tuple(list_value.items())
                     list_values.append(dict_as_tuple)
                 else:
                     list_values.append(list_value)
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -201,7 +201,27 @@ class Record:
         return result
 
     def __hash__(self) -> int:
-        return hash(self._pack(excluded_fields=IGNORE_FIELDS_FOR_COMPARISON))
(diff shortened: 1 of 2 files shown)


flow/record/base.py
```diff
@@ -201,7 +201,27 @@ def _replace(self, **kwds):
...
+        for value in values:
+            if not isinstance(value, list):
+                record_values.append(value)
+                continue
+            list_values = []
+            for list_value in value:
+                if isinstance(list_value, dict):
+                    # List values that are dicts must be converted to tuples
+                    dict_as_tuple = ((k, v) for k, v in list_value.items())
+                    list_values.append(dict_as_tuple)
```
yunzheng, line 219: Can be simplified to:

```suggestion
                    list_values.append(tuple(list_value.items()))
```

tests/test_record.py
```diff
@@ -922,3 +922,16 @@ def test_ignore_fields_for_comparision_contextmanager():
...
+def test_list_field_type_hashing():
+    TestRecord = RecordDescriptor(
+        "test/record",
+        [
+            ("string[]", "stringlist"),
+            ("dictlist", "dictlist"),
+        ],
+    )
+
+    test_record = TestRecord(stringlist=["a", "b", "c", "d"], dictlist=[{"a": "b", "c": "d"}])
```
yunzheng, line 936: Add one extra dictonary item to dictlist just in case:

```suggestion
    test_record = TestRecord(stringlist=["a", "b", "c", "d"], dictlist=[{"a": "b", "c": "d"}, {"foo": "bar"}])
```

yunzheng (changes requested):

yunzheng (approved):

PR #120, opened 2024-04-22
Empty path str() should not return cwd

cecinestpasunepipe:
(DIS-2557)

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -734,3 +734,31 @@ class windows_path(pathlib.PureWindowsPath, path):
                 quote = '"'
 
         return f"{quote}{s}{quote}"
+
+
+# DIS-2557: str(Path("")) should not return "."
+if not hasattr(pathlib.PurePath, "__old_str__"):
+    if PY_312:
+
+        def _new__str__(self) -> str:
+            return "" if "".join(self._raw_paths).strip() == "" else pathlib.PurePath.__old_str__(self)
+
+    else:
+
+        def _new_parse_parts(self, parts: list[str | pathlib.PurePath]) -> tuple(str, str, str):
+            raw_path = "".join(parts).strip()
+            self.empty = raw_path == ""
+            drv, root, parsed = pathlib._Flavour._old_parse_parts(self, parts)
+            # flavour removes . from parts, so we need to add it back otherwise path(pathlib.Path(".")) == ""
+            if raw_path == ".":
+                parsed.append(".")
+            return drv, root, parsed
+
+        def _new__str__(self) -> str:
+            return "" if self._flavour.empty else pathlib.PurePath.__old_str__(self)
+
+        pathlib._Flavour._old_parse_parts = pathlib._Flavour.parse_parts
+        pathlib._Flavour.parse_parts = _new_parse_parts
+
+    pathlib.PurePath.__old_str__ = pathlib.PurePath.__str__
+    pathlib.PurePath.__str__ = _new__str__
--- a/tests/test_fieldtypes.py
+++ b/tests/test_fieldtypes.py
@@ -589,6 +589,41 @@ def test__is_windowslike_path(path_, is_windows):
     assert _is_windowslike_path(path_) == is_windows
 
 
+@pytest.mark.parametrize(
+    "path_, expected_str",
+    [
+        (None, None),
+        (".", "."),
+        ("", ""),
+        (" ", ""),
+        ("  ", ""),
+        ("\t", ""),
+        ("\t\t", ""),
+        ("\r", ""),
+        ("\r\r", ""),
+        ("\n", ""),
+        ("\n\n", ""),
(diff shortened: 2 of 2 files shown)


flow/record/fieldtypes/__init__.py
```diff
@@ -734,3 +734,26 @@ def __repr__(self) -> str:
...
+            return pathlib._Flavour._old_parse_parts(self, parts)
+
+        def _new__str__(self) -> str:
+            return "" if self._flavour.empty else pathlib.PurePath.__old_str__(self)
+
+        pathlib._Flavour._old_parse_parts = pathlib._Flavour.parse_parts
+        pathlib._Flavour.parse_parts = _new_parse_parts
+
+    pathlib.PurePath.__old_str__ = pathlib.PurePath.__str__
+    pathlib.PurePath.__str__ = _new__str__
```
Schamper, line 759: Wouldn't this change the behaviour of _all_ `Path` objects as a side effect?

```python
In [1]: from flow.record import fieldtypes

In [2]: from pathlib import Path

In [3]: Path("")
Out[3]: PosixPath('')
```

Schamper (changes requested):

flow/record/fieldtypes/__init__.py
cecinestpasunepipe: Otherwise:

`path(Path(""))`

will still give you `"."`

Maybe I am missing something but to make this work I think I need to change the default behavior of Path, otherwise the data is lost. Or we have to replace every instance of Path with a different class, like ForensicPath (that keeps the original args)?

tests/test_fieldtypes.py
```diff
@@ -589,6 +589,41 @@ def test__is_windowslike_path(path_, is_windows):
...
+def test_empty_path_str(path_: str, expected_str: str) -> None:
+    TestRecord = RecordDescriptor(
+        "test/path",
+        [
+            ("path", "value"),
+        ],
+    )
+
+    r = TestRecord(path_)
+    if path_ is None:
```
yunzheng, line 617: it would be better to just move `None` to it's own test to simplify this test.

tests/test_fieldtypes.py
```diff
@@ -589,6 +589,41 @@ def test__is_windowslike_path(path_, is_windows):
...
+        (".", "."),
+        ("", ""),
+        (" ", ""),
+        ("  ", ""),
+        ("\t", ""),
+        ("\t\t", ""),
+        ("\r", ""),
+        ("\r\r", ""),
+        ("\n", ""),
+        ("\n\n", ""),
```
yunzheng, line 605: The test is incorrect as these can all be valid paths and should not be stripped:

```python
In [1]: from pathlib import Path

In [2]: open("\t\t", "w").write("tab tab")
Out[2]: 7

In [3]: str(Path("\t\t"))
Out[3]: '\t\t'

In [4]: Path("\t\t").read_text()
Out[4]: 'tab tab'
```

yunzheng (commented):
The title of the PR is a bit misleading as `cwd` is not resolved to the current working directory but just a dot.

Do you have any examples where this is an issue and is considered a bug? Maybe in most cases if you encounter an empty string as a path value, setting the record value to `None` would be better.

Schamper:
> Do you have any examples where this is an issue and is considered a bug? Maybe in most cases if you encounter an empty string as a path value, setting the record value to `None` would be better.

There can be cases where we parse some artefact which may or may not have a path field. IMO there's a difference between the absence of the path field (`None`) or an empty path field (`""`) and thus it should be possible to be noted as such in the record.

yunzheng:
My take on this issue: https://github.com/fox-it/flow.record/pull/122

As this is a totally different way to tackle this problem from this PR i've created a new branch and PR.

tests/test_fieldtypes.py
cecinestpasunepipe: Ok, sounds logical. I was a bit unsure about this, so I awaited your feedback regarding the matter.

cecinestpasunepipe:
> My take on this issue: #122
> 
> As this is a totally different way to tackle this problem from this PR i've created a new branch and PR.

I added some comments to your PR. I like your solution and I was thinking along the same lines, however like I pointed out in my other comment, it does not seem to meet the requested requirements.

flow/record/fieldtypes/__init__.py
yunzheng: I don't think we want to change default behaviour of stdlib classes.


> path(Path(""))
> 
> will still give you "."
> 

I think the above is "expected" behaviour as stdlib Path does not support empty paths, but we do want to support it in our own path objects.

yunzheng:
I left a comment in the first unresolved thread with my point of view, my other suggestions are still unresolved. If you want me to review anything else let me know.

yunzheng:
After some discussion, this PR is closed in favour of https://github.com/fox-it/flow.record/pull/122

PR #122, opened 2024-05-07
Add support for empty value in path fieldtype

yunzheng:
Normally an empty path would be normalized to a "." (dot) character.
This change allows you to initialize a path field with an empty string.
This is useful to represent a path that is empty.

Fixes DIS-2557

### DIFF
--- a/tests/test_fieldtypes.py
+++ b/tests/test_fieldtypes.py
@@ -1146,14 +1146,20 @@ def test_empty_path(path_cls) -> None:
     assert p1 == ""
     assert p1._empty_path
     assert str(p1) == ""
+    assert p1 != path_cls(".")
 
     # initialize without any arguments
     p2 = path_cls()
     assert p2 == ""
     assert p2._empty_path
     assert str(p2) == ""
+    assert p2 != path_cls(".")
 
     assert p1 == p2
+
+
+def test_empty_path_different_types() -> None:
+    assert fieldtypes.posix_path("") != fieldtypes.windows_path("")
 
 
 def test_record_empty_path() -> None:
--- a/tests/test_fieldtypes.py
+++ b/tests/test_fieldtypes.py
@@ -11,7 +11,7 @@ from datetime import datetime, timedelta, timezone
 import pytest
 
 import flow.record.fieldtypes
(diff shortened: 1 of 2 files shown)


yunzheng:
Currently `flow.record.fieldtypes.path()` (without any arguments) will also be considered an empty path. Not sure if this is preferred or to only consider an empty path if it's initialized with an empty string `""`.

yunzheng:
No idea why Python 3.12 fails, will look into it later

tests/test_fieldtypes.py
```diff
@@ -1132,5 +1132,79 @@ def test_command_failed() -> None:
...
+    ],
+)
+def test_empty_path(path_cls) -> None:
+    # initialize with empty string
+    p1 = path_cls("")
+    assert p1 == ""
+    assert p1._empty_path
+    assert str(p1) == ""
+    assert p1 != path_cls(".")
+
```
cecinestpasunepipe, line 1150: Your solution is somewhat similar to my original proposal, which was rejected because it failed the following test (which also fails here):

```
 # initialize with empty path
    p1 = path_cls(Path(""))
    assert p1 == ""
    assert p1._empty_path
    assert str(p1) == ""
    assert p1 != path_cls(".")
```

cecinestpasunepipe (changes requested):

tests/test_fieldtypes.py
cecinestpasunepipe: According to my information this was part of the requirements. I see no other way to implement it.

Schamper: As long as this works with our own `TargetPath`, all should be well in the world.

yunzheng: There was not much information regarding the requirements in the DIS ticket, other than that Record path fields do not support empty values as they would end up as the `.` dot value. This PR fixes that.

If `windows_path` and `posix_path` fields are intitialised using `TargetPath` variables, then we need to ensure that `TargetPath` also supports empty paths and then we can test if that works.

cecinestpasunepipe (approved):

PR #123, opened 2024-05-08
Fix ValueError: I/O operation on closed file during tests

yunzheng:
Sometimes the `stdout` file object is closed by `flow.record` internals as it is sometimes mocked and swapped by `pytest` during tests which in turn can confuse `is_stdout()` to return `False` causing the file to be closed.

This is now fixed by adding two custom methods for getting the stdio streams:

 * `flow.record.utils.get_stdout()`
 * `flow.record.utils.get_stdin()`

These methods are the preferred way to get the stdio streams as they also set an extra attribute on the returned file object that is checked by `is_stdout()`.


yunzheng:
This sadly didn't fix the issue :(

Will leave this PR in Draft state for more adhoc tests.

yunzheng:
Took a while but this should now be fixed. I took quite a deep dive into the pytest internals to figure out the cause.

It's mainly caused due to pytest's capture.py swapping `sys.stdout` with it's own objects, however in our own code we check if the file object is `sys.stdout` or `sys.stdout.buffer`, but as `sys.stdout` keeps getting swapped multiple times (eg: nested pytest operations/fixtures) it could be that a file object that is originally stdout is now not seen as stdout anymore, and therefore closed by `flow.record` internals at certains points.

flow/record/utils.py
```diff
@@ -1,15 +1,49 @@
...
 
-def is_stdout(fp):
-    return fp in (sys.stdout, sys.stdout.buffer)
+def get_stdout(binary: bool = False) -> TextIO | BinaryIO:
+    """Return the stdout stream as binary or text stream.
+
+    This function is the preferred way to get the stdout stream in flow.record.
+
+    Arguments:
+        binary: Whether to return the stream as binary stream.
```
Schamper, line 20:
```suggestion
        binary: Whether to return the stream as binary stream.

```

flow/record/utils.py
```diff
@@ -1,15 +1,49 @@
...
+    return fp
+
+
+def get_stdin(binary: bool = False) -> TextIO | BinaryIO:
+    """Return the stdin stream as binary or text stream.
+
+    This function is the preferred way to get the stdin stream in flow.record.
+
+    Arguments:
+        binary: Whether to return the stream as binary stream.
```
Schamper, line 35:
```suggestion
        binary: Whether to return the stream as binary stream.

```

flow/record/utils.py
```diff
@@ -1,15 +1,49 @@
...
+    Returns:
+        The stdin stream.
+    """
+    fp = getattr(sys.stdin, "buffer", sys.stdin) if binary else sys.stdin
+    fp._is_stdin = True
+    return fp
+
+
+def is_stdout(fp: TextIO | BinaryIO) -> bool:
+    """Returns True if `fp` is the stdout stream."""
```
Schamper, line 45:
```suggestion
    """Returns True if ``fp`` is the stdout stream."""
```

Schamper (changes requested):

PR #124, opened 2024-05-27
Use non-dev dependencies in pyproject.toml

narimantos:
https://github.com/fox-it/dissect/issues/61

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,13 @@ deps =
     pytest-cov
     coverage
 extras = test
+# Unfortunately, tox does not allow separate installation flags for the project
+# dependencies and the test dependencies. When running tox, we want to install the
+# project dependencies with the --pre flag, so that we get the latest version of all
+# dependencies. We do the installation step ourselves for this reason.
+skip_install = true
+commands_pre =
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -16,6 +16,13 @@ deps =
     pytest-cov
     coverage
 extras = test
```
yunzheng, line 18: i think the extras can be removed here then?

tox.ini
```diff
@@ -16,6 +16,13 @@ deps =
...
+# Unfortunately, tox does not allow separate installation flags for the project
+# dependencies and the test dependencies. When running tox, we want to install the
+# project dependencies with the --pre flag, so that we get the latest version of all
+# dependencies. We do the installation step ourselves for this reason.
+skip_install = true
+commands_pre =
+    pip install --pre -e .[test]
```
yunzheng, line 25: It's also possible to list the flags and package name under `[deps]` like so:

```
[testenv]
deps =
    pytest
    pytest-cov
    coverage
    --pre --editable .[test]
```

Then you don't have to run the `pip` command separately.

PR #129, opened 2024-08-01
Fix JsonRecordPacker boolean packing

JSCU-CNI:
This PR fixes `boolean` field type packing behaviour of the `JsonRecordPacker`. Currently record types of `boolean` are not converted to JSON booleans but become integers. This patch fixes that.

### DIFF
--- a/flow/record/jsonpacker.py
+++ b/flow/record/jsonpacker.py
@@ -51,7 +51,7 @@ class JsonRecordPacker:
                 if field_type == "bytes" and isinstance(serial[field_name], str):
                     serial[field_name] = base64.b64encode(serial[field_name]).decode()
 
-                # Boolean field types should be case to a bool instead of staying ints
+                # Boolean field types should be cast to a bool instead of staying ints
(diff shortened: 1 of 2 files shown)


flow/record/jsonpacker.py
```diff
@@ -41,15 +41,20 @@ def pack_obj(self, obj):
...
                 serial["_type"] = "record"
                 serial["_recorddescriptor"] = obj._desc.identifier
 
-            # PYTHON2: Because "bytes" are also "str" we have to handle this here
             for field_type, field_name in obj._desc.get_field_tuples():
+                # PYTHON2: Because "bytes" are also "str" we have to handle this here
                 if field_type == "bytes" and isinstance(serial[field_name], str):
                     serial[field_name] = base64.b64encode(serial[field_name]).decode()
 
+                # Boolean field types should be case to a bool instead of staying ints
```
Schamper, line 54:
```suggestion
                # Boolean field types should be cast to a bool instead of staying ints
```

Schamper (changes requested):

Schamper (approved):

PR #130, opened 2024-08-05
Flatten command field types for the jsonpacker

### LINKED ISSUE #132: Flatten command field types for the jsonpacker

JSCU-CNI:
This PR flattens the field type `command` in the JSON packer and fixes #132.

Currently the `dissect.target` project is inconsistent in using the same field **name** `command` and the new field *type* `command`. This patch makes it possible to upload and aggregate on different records in Elasticsearch with the field name `command` and differing field types.

For example, see `RunKeysPlugin.runkeys` and `PowerShellHistoryPlugin.powershell_history`.

You could argue (and we agree) that this should be fixed in `dissect.target` as all `RecordDescriptor`s currently using `("string", "command")` should perhaps use the new `command ` record type. That makes sense to do in the long run. Perhaps a field called `full` could be added to the standard output of the `command` fieldtype dict to still be able to index the full, original, command.

Historically the `command` field type introduced a backwards incompatible change into dissect. This PR fixes that inconsistency.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,6 @@
+# Formatting commits. You can ignore them during git-blame with `--ignore-rev` or `--ignore-revs-file`.
+#
+#   $ git config --add 'blame.ignoreRevsFile' '.git-blame-ignore-revs'
+#
+# Change linter to Ruff (#158)
+c67f778c653c295ec26146cf6422d3b06ac640e8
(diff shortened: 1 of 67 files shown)


gardianofthisgalaxy (approved):

EinatFox:
Hi @JSCU-CNI , 
This PR is on our radar but unfortunately due to low capacity I cannot yet tell when it will get priority. 
However, we do agree about the need to fix the inconsistency that record fields named command should not use the command type. I have created a separate issue for that in dissect.target #984. If you can make a PR for it, it will get prioritized for review, and hopefully this will partially resolve your issue.

JSCU-CNI:
Superseded by #183.

PR #137, opened 2024-09-05
Improve XLSX adapter

JSCU-CNI:
This PR makes it easier to write records to an xlsx file. Each record descriptor gets its own sheet within the workbook and field values are sanitized to prevent openpyxl errors.

### DIFF
--- a/flow/record/adapter/xlsx.py
+++ b/flow/record/adapter/xlsx.py
@@ -1,5 +1,5 @@
 from base64 import b64decode, b64encode
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any, Iterator
 
 from openpyxl import Workbook, load_workbook
@@ -16,7 +16,7 @@ __usage__ = """
 Microsoft Excel spreadsheet adapter
 ---
 Write usage: rdump -w xlsx://[PATH]
-Read usage:  rdump xlsx://[PATH]
+Read usage: rdump xlsx://[PATH]
 [PATH]: path to file. Leave empty or "-" to output to stdout
 """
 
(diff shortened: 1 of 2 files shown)


flow/record/adapter/xlsx.py
```diff
@@ -1,36 +1,94 @@
...
+def sanitize_fieldvalues(values: Iterator[Any]) -> Iterator[Any]:
+    """Sanitize field values so openpyxl will accept them."""
+
+    for value in values:
+        # openpyxl doesn't support timezone-aware datetime instances,
+        # so we convert to UTC and then remove the timezone info.
+        if isinstance(value, datetime) and value.tzinfo is not None:
+            utc_offset = value.utcoffset()
+            value += utc_offset
+            value = value.replace(tzinfo=None)
```
yunzheng, line 33: this should be `value -= utc_offset` or preferably use `.astimezone()`, eg:

```python
from datetime import datetime, timezone

value = value.astimezone(timezone.utc).replace(tzinfo=None)
```

tests/test_xlsx_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+                ["Shaken", "Not", "Stirred"],
+                fieldtypes.posix_path("/home/user"),
+                fieldtypes.posix_command("/bin/bash -c 'echo hello world'"),
+                fieldtypes.windows_path("C:\\Users\\user\\Desktop"),
+                fieldtypes.windows_command("C:\\Some.exe /?"),
+            ]
+        )
+    ) == [
+        7,
+        datetime(1920, 11, 11, 15, 37, 0),  # UTC normalization
```
yunzheng, line 45:
```suggestion
        datetime(1920, 11, 11, 11, 37, 0),  # UTC normalization
```

yunzheng (changes requested):

yunzheng (approved):

PR #145, opened 2024-10-01
Fix the build of cramjam a subdependency of fastavro

pyrco:

### DIFF
--- a/.github/workflows/dissect-ci.yml
+++ b/.github/workflows/dissect-ci.yml
@@ -11,8 +11,6 @@ on:
 jobs:
   ci:
     uses: fox-it/dissect-workflow-templates/.github/workflows/dissect-ci-template.yml@main
-    with:
-      deb-packages: nasm
     secrets: inherit
 
   publish:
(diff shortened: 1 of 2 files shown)


pyrco:
This fixes it for pyp3.9 on Linux. The windows fail seems rather more complicated. We could make that build not required anymore?

yunzheng (commented):
this sadly doesn't fix the windows pypy3.9. Maybe we should pin cramjam to an older version for now?

yunzheng:
looks like 3.9 support was removed from cramjam. See https://github.com/milesgranger/cramjam/pull/177

yunzheng:
@pyrco I've pinned cramjam for now for pypy3.9, which seems to be the easiest to do for now. 

I think you can also remove the extra deb package now. But i leave that up to you.

pyrco:
Let's go for the pinning option and see when that fails ;-) If it does there are some other options, which would disable avro support in flow.record on pypy3.9 & windows.

In the meantime we should look into dropping support for pypy3.9. We want to keep CPython 3.9 around for a bit though. The current idea is to stop supporting that once it is out of support upstream.

yunzheng (approved):

PR #147, opened 2024-10-08
Add assetions  for explicit path comparison.

twiggler:
Originally, this issue was about implementing explicit path comparisons such as `path == posix_path(/usr/local/ls").
As it turns out, this behavior is already supported by flow.record `path` because it delegates to the superclass method `pathlib::__eq__`.

Added some assertions (actually I started with these and found out it already works as intended). Kept them to prevent  regressions.

Bonus: improve typing of tests.


tests/test_fieldtypes.py
```diff
@@ -452,7 +453,7 @@ def test_datetime():
         ("2006-11-10T14:29:55.585192699999999-07:00", datetime(2006, 11, 10, 21, 29, 55, 585192, tzinfo=UTC)),
     ],
 )
-def test_datetime_formats(tmp_path, value, expected_dt):
+def test_datetime_formats(tmp_path, value, expected_dt) -> None:
```
Schamper, line 456: Could you also add types to the arguments?

Schamper (approved):

Schamper (changes requested):
Wrong button

tests/test_fieldtypes.py
twiggler: Added.

tests/test_fieldtypes.py
```diff
@@ -689,7 +690,9 @@ def test_path() -> None:
         ),
     ],
 )
-def test_path_multiple_parts(path_parts, expected_instance) -> None:
+def test_path_multiple_parts(
+    path_parts: Tuple[str | pathlib.PurePath, ...], expected_instance: Type[pathlib.PurePath]
```
Schamper, line 694:
```suggestion
    path_parts: Tuple[str | pathlib.PurePath, ...], expected_instance: type[pathlib.PurePath]
```
This is legal these days (unless it's another thing not allowed by 3.8, in which case ignore this comment).

tests/test_fieldtypes.py
```diff
@@ -1153,7 +1160,7 @@ def test_command_failed() -> None:
         fieldtypes.path,
     ],
 )
-def test_empty_path(path_cls) -> None:
+def test_empty_path(path_cls: Type[pathlib.PurePath]) -> None:
```
Schamper, line 1163:
```suggestion
def test_empty_path(path_cls: type[pathlib.PurePath]) -> None:
```

Schamper (changes requested):

tests/test_fieldtypes.py
twiggler: I thought this got introduced in python 3.9 but the python compatibility test passes with `type` instead of `Type`

Schamper (dismissed):

tests/test_fieldtypes.py
```diff
@@ -688,7 +690,9 @@ def test_path():
         ),
     ],
 )
-def test_path_multiple_parts(path_parts, expected_instance):
+def test_path_multiple_parts(
+    path_parts: Tuple[str | pathlib.PurePath, ...], expected_instance: type[pathlib.PurePath]
```
pyrco, line 694:
```suggestion
    path_parts: tuple[str | pathlib.PurePath, ...], expected_instance: type[pathlib.PurePath]
```

pyrco (changes requested):

pyrco (approved):

PR #150, opened 2024-10-31
Improve elastic adapter exception handling

JSCU-CNI:
This PR improves exception handling of the Elastic adapter. Currently when an Elasticsearch endpoint is unreachable `rdump` will stay stuck in a thread. We have added a `threading.excepthook` to make sure `rdump` exists with a non-zero status code. Let us know if you prefer a different method of propagating exceptions in rdump. Currently the stdout is a bit noisey, but at least rdump is no longer stuck.

Steps to reproduce:

```
rdump foo.rec -w elastic+https://invalid-hostname?index=foo
```

### DIFF
--- a/flow/record/adapter/elastic.py
+++ b/flow/record/adapter/elastic.py
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 import hashlib
 import logging
 import queue
 import threading
-from typing import Iterator, Optional, Union
+from typing import Iterator
 
 import elasticsearch
 import elasticsearch.helpers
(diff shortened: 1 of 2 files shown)


flow/record/adapter/elastic.py
```diff
@@ -138,21 +149,24 @@ def flush(self) -> None:
         pass
 
     def close(self) -> None:
+        self.queue.put(StopIteration)
+        self.event.wait()
```
yunzheng, line 153: note that `self.queue` and `self.event` here might not exist yet if our `__init__` raises an exception, which results that this `close()` method is called due to a finally block and causing an AttributeError as they are not initialised yet.

That was the reason why we added `if hasattr(self, "es")` guard, which in hindsight is not very clean.

I think it would be safe to move `self.queue = ...` and `self.event = ...` higher up (before `self.es = ...`) in the `__init__` and it would be OK.

yunzheng (changes requested):

yunzheng (approved):

PR #154, opened 2024-12-04
Add path compatibility with Python 3.13

### LINKED ISSUE #153: Support Python 3.13

Schamper:
Fixes #153

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -33,6 +33,7 @@ UTC = timezone.utc
 
 PY_311_OR_HIGHER = sys.version_info >= (3, 11, 0)
 PY_312_OR_HIGHER = sys.version_info >= (3, 12, 0)
+PY_313_OR_HIGHER = sys.version_info >= (3, 13, 0)
 
 TYPE_POSIX = 0
 TYPE_WINDOWS = 1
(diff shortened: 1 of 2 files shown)


flow/record/fieldtypes/__init__.py
```diff
@@ -600,12 +601,18 @@ def _unpack(cls, data):
         return data
 
 
-def _is_posixlike_path(path: Any):
-    return isinstance(path, pathlib.PurePath) and "\\" not in (path._flavour.sep, path._flavour.altsep)
+def _is_posixlike_path(path: Any) -> bool:
+    if isinstance(path, pathlib.PurePath):
+        obj = getattr(path, "parser", None) or path._flavour
```
yunzheng, line 606: Would this also work?

```suggestion
        obj = getattr(path, "parser", path._flavour)
```

flow/record/fieldtypes/__init__.py
```diff
@@ -600,12 +601,18 @@ def _unpack(cls, data):
...
+        return "\\" not in (obj.sep, obj.altsep)
+    return False
 
 
-def _is_windowslike_path(path: Any):
-    return isinstance(path, pathlib.PurePath) and "\\" in (path._flavour.sep, path._flavour.altsep)
+def _is_windowslike_path(path: Any) -> bool:
+    if isinstance(path, pathlib.PurePath):
+        obj = getattr(path, "parser", None) or path._flavour
```
yunzheng, line 613:
```suggestion
        obj = getattr(path, "parser", path._flavour)
```

yunzheng (changes requested):

flow/record/fieldtypes/__init__.py
Schamper: No, since `._flavour` will result in an attribute error on newer Python versions because it'll get eagerly executed to pass the result to `getattr`. In this syntax, `._flavour` only gets executed in case `parser` does not exist (happy path for the future).

yunzheng (approved):

PR #155, opened 2025-01-08
Handle NoneType in elastic adapter document_stream

JSCU-CNI:
This PR fixes a `Unpackable type <class 'NoneType'>` exception for the elastic adapter. This can occur when the record queue gets populated with `None` instead of a record - which apparently can occur if for example a Dissect plugin does not return any records.


yunzheng:
Do you know which dissect plugin by any chance? It sounds like this should also be fixed there.

PR #156, opened 2025-01-08
Add queue size to Elastic adapter

raverburg:
Add queue size limit to prevent high memory usage for plugins with multiple million of records

During testing the elastic adapter memory usage was very high (50GB+) for plugins such as MFT and UsnJrnl with high amounts of records, this fix limits the memory usage by adding a max size to the queue.

### DIFF
--- a/flow/record/adapter/elastic.py
+++ b/flow/record/adapter/elastic.py
@@ -54,7 +54,7 @@ class ElasticWriter(AbstractWriter):
         if not uri.lower().startswith(("http://", "https://")):
             uri = "http://" + uri
 
-        self.queue: queue.Queue[Record | StopIteration] = queue.Queue()
+        self.queue: queue.Queue[Record | StopIteration] = queue.Queue(maxsize=100000)
         self.event = threading.Event()
 
         self.es = elasticsearch.Elasticsearch(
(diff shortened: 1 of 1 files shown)


flow/record/adapter/elastic.py
```diff
@@ -25,6 +25,7 @@
 
 Optional arguments:
   [API_KEY]: base64 encoded api key to authenticate with (default: False)
+  [QUEUE_SIZE]: integer that sets the limit on the maximum number of records in the queue. Limits memory usage (default: 100000)
```
yunzheng, line 28: Made the description a bit more concise:
```suggestion
  [QUEUE_SIZE]: maximum queue size for writing records; limits memory usage (default: 100000)
```

yunzheng (changes requested):

flow/record/adapter/elastic.py
```diff
@@ -43,18 +44,20 @@ def __init__(
         http_compress: str | bool = True,
         hash_record: str | bool = False,
         api_key: str | None = None,
+        queue_size: int = 100000,
```
EdoHooghiemster, line 47: Magic number. U should put this in a const

yunzheng: While using constants for magic numbers can be good practice, in this context the value is straightforward enough that it doesn't significantly impact readability.

yunzheng (approved):

yunzheng:
Thanks for the PR, it's been merged!

PR #160, opened 2025-02-06
Add ipinterface type to fieldtypes.net

Miauwkeru:

### DIFF
--- a/flow/record/fieldtypes/net/ip.py
+++ b/flow/record/fieldtypes/net/ip.py
@@ -20,7 +20,7 @@ _IPNetwork = Union[IPv4Network, IPv6Network]
 _IPAddress = Union[IPv4Address, IPv6Address]
 _IPInterface = Union[IPv4Interface, IPv6Interface]
 _ConversionTypes = Union[str, int, bytes]
-_IP = Union[_IPNetwork, _IPAddress, _IPInterface]
+_IPTypes = Union[_IPNetwork, _IPAddress, _IPInterface]
 
 
 class ipaddress(FieldType):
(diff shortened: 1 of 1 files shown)


flow/record/fieldtypes/net/ip.py
```diff
@@ -98,8 +102,56 @@ def _pack(self) -> str:
...
+    def ip(self) -> ipaddress:
+        return ipaddress(self.val.ip)
+
+    @property
+    def network(self) -> ipnetwork:
+        return ipnetwork(self.val.network)
+
+    @property
+    def netmask(self) -> ipnetwork:
+        return self.network.netmask
```
yunzheng, line 142: i believe this should return a `ipaddress`, can you double check?

flow/record/jsonpacker.py
```diff
@@ -79,6 +79,12 @@ def pack_obj(self, obj: Any) -> dict | str:
                 "executable": obj.executable,
                 "args": obj.args,
             }
+        if isinstance(obj, fieldtypes.net.ipinterface):
+            return {
+                "ip": str(obj.ip),
+                "subnetmask": str(obj.netmask),
+                "network": str(obj.network),
+            }
```
yunzheng, line 87: is returning a whole dictionary necessary? i would prefer to keep it simple and just use the string representation of the `ipinterface` object, eg: `192.168.1.10/24`.

flow/record/fieldtypes/net/ip.py
```diff
@@ -15,6 +18,7 @@
 
 _IPNetwork = Union[IPv4Network, IPv6Network]
 _IPAddress = Union[IPv4Address, IPv6Address]
+_IPInterface = Union[IPv4Interface, IPv6Interface]
```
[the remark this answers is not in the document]
yunzheng: Looks like this indeed requires Python 3.10+ for it to work, even with `from __future__ import annotations` in Python 3.9.

flow/record/fieldtypes/net/ip.py
```diff
@@ -15,16 +18,19 @@
...
+_ConversionTypes = Union[str, int, bytes]
+_IP = Union[_IPNetwork, _IPAddress, _IPInterface]
```
yunzheng, line 23: Can you rename `_IP` to `_IPTypes`, that will be clearer and more consistent.

yunzheng (changes requested):

yunzheng (approved):

PR #164, opened 2025-02-17
Better exception handling for Elastic adapter

JSCU-CNI:
This PR fixes two issues with the Elastic adapter. Currently the Elastic adapter fails silently  - as in it does not break but continues - when a record is not inserted in the index. We make sure to raise a `ValueError` instead so the `threading.excepthook` catches the error and stops the elastic writer.

The second issue we fix here is we make sure both reader and writer threads crash when `self.exception` is filled. This prevents a hanging reader thread if the writer thread fails.

### DIFF
--- a/flow/record/adapter/elastic.py
+++ b/flow/record/adapter/elastic.py
@@ -15,14 +15,6 @@ try:
 except ImportError:
     HAS_ELASTIC = False
 
-try:
-    import tqdm
-
-    HAS_TQDM = True
-
-except ImportError:
-    HAS_TQDM = False
-
 from flow.record.adapter import AbstractReader, AbstractWriter
 from flow.record.base import Record, RecordDescriptor
 from flow.record.fieldtypes import fieldtype_for_value
@@ -119,11 +111,6 @@ class ElasticWriter(AbstractWriter):
             if arg_key.startswith("_meta_"):
                 self.metadata_fields[arg_key[6:]] = arg_val
 
-        if HAS_TQDM:
-            self.progress = tqdm.tqdm(unit=" records", delay=3)
-        else:
-            self.progress = None
-
     def excepthook(self, exc: threading.ExceptHookArgs, *args, **kwargs) -> None:
         log.error("Exception in thread: %s", exc)
         self.exception = getattr(exc, "exc_value", exc)
@@ -189,8 +176,7 @@ class ElasticWriter(AbstractWriter):
             # Some settings have to be redefined because streaming_bulk does not inherit them from the self.es instance.
             max_retries=3,
         ):
-            if HAS_TQDM:
-                self.progress.update(1)
+            pass
 
         self.event.set()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ compression = [
 ]
 elastic = [
     "elasticsearch",
-    "tqdm",
 ]
 geoip = [
     "maxminddb",


flow/record/adapter/elastic.py
```diff
@@ -149,11 +149,14 @@ def streaming_bulk_thread(self) -> None:
             raise_on_exception=False,
```
yunzheng, line 149: I wonder if we cannot just set `raise_on_error=True` and  `raise_on_exception=True` in `streaming_bulk`, would that also work?

yunzheng (changes requested):
next to the inline comment, can you also move the following code block:

```python
 self.exception: Exception | None = None
 threading.excepthook = self.excepthook
```

to before `self.es = elasticsearch.Elasticsearch()`.

This ensures that if it fails during instantiating `ElasticSearch()`, that `self.exception` is available and does not cause another exception during `close()`.

Schamper:
> We added a simple pure python progress bar for the elastic writer in [06cfa15](https://github.com/fox-it/flow.record/pull/164/commits/06cfa152eb9dc198ee0cdd02eea7af72dd6341d3). `tqdm` is a minimal and fast dependency and allows one to see how many records have been processed by rdump in a TTY. 
> 
> 
> 
> ```
> 
> $ target-query -q -f mft example.img | rdump -w "elastic+http://foo:bar@localhost:9200?verify_certs=0&index=test"
> 
> [reading from stdin]
> 
> 2199501 records [06:19, 5041.39 records/s]
> 
> ```

On the topic of progress bars, I thought of a mechanism recently that will generically allow any `dissect.target` plugin to provide progress "feedback", making it possible to provide actual completion progress bars for plugins where that info is calculatable, and more generic progress bars for plugins where it isn't.

I plan to write down these concepts in a few issues next week.

yunzheng:
> > We added a simple pure python progress bar for the elastic writer in [06cfa15](https://github.com/fox-it/flow.record/pull/164/commits/06cfa152eb9dc198ee0cdd02eea7af72dd6341d3). `tqdm` is a minimal and fast dependency and allows one to see how many records have been processed by rdump in a TTY.
> > ```
> > 
> > $ target-query -q -f mft example.img | rdump -w "elastic+http://foo:bar@localhost:9200?verify_certs=0&index=test"
> > 
> > [reading from stdin]
> > 
> > 2199501 records [06:19, 5041.39 records/s]
> > ```
> 
> On the topic of progress bars, I thought of a mechanism recently that will generically allow any `dissect.target` plugin to provide progress "feedback", making it possible to provide actual completion progress bars for plugins where that info is calculatable, and more generic progress bars for plugins where it isn't.
> 
> I plan to write down these concepts in a few issues next week.

I ❤️ progress bars.

To comment on this from the `flow.record` perspective, I've always wanted to add progress bar support and I think it's best if we split the tqdm/progress bar support in the elastic adapter into a separate issue and add this to `tools/rdump.py` so all adapters can benefit from it.

flow/record/adapter/elastic.py
```diff
@@ -140,20 +175,29 @@ def document_stream(self) -> Iterator[dict]:
...
     def streaming_bulk_thread(self) -> None:
-        """Thread that streams the documents to ES via the bulk api"""
+        """Thread that streams the documents to ES via the bulk api.
 
-        for ok, item in elasticsearch.helpers.streaming_bulk(
+        Resources:
+            - https://elasticsearch-py.readthedocs.io/en/v8.17.1/helpers.html#elasticsearch.helpers.streaming_bulk
+            - https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/helpers/actions.py#L362
+        """
+        for _, _ in elasticsearch.helpers.streaming_bulk(
```
yunzheng, line 184:
```suggestion
        for _ok, _item in elasticsearch.helpers.streaming_bulk(
```

flow/record/adapter/elastic.py
```diff
@@ -49,8 +65,19 @@ def __init__(
...
         **kwargs,
     ) -> None:
+        """Initialize the ElasticWriter.
+
+        Resources:
+            - https://elasticsearch-py.readthedocs.io/en/v8.17.1/api/elasticsearch.html
+        """
+
+        if not HAS_ELASTIC:
+            raise ValueError("Required dependency elasticsearch missing")
```
yunzheng, line 79:
```suggestion
        if not HAS_ELASTIC:
            raise RuntimeError("Required dependency 'elasticsearch' missing")
```

yunzheng (changes requested):
It would be great if the tqdm code can be moved and incorporated in a new PR. See https://github.com/fox-it/flow.record/issues/165

yunzheng (approved):

PR #170, opened 2025-06-17
Add -Cn/--csv-no-header option to rdump

yunzheng:
This allows you to dump the fields as CSV but without the row header.

Also fixes issue that in some cases the selected fields were not propagated to the adapter.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -17,5 +17,9 @@ def test_boolean_argument() -> None:
     assert boolean_argument("n") is False
     assert boolean_argument("on") is True
     assert boolean_argument("off") is False
+    assert boolean_argument(True) is True
+    assert boolean_argument(False) is False
+    assert boolean_argument(1) is True
+    assert boolean_argument(0) is False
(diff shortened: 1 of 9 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,21 @@
...
+    assert boolean_argument("False") is False
+    assert boolean_argument("false") is False
+    assert boolean_argument("1") is True
+    assert boolean_argument("0") is False
+    assert boolean_argument("yes") is True
+    assert boolean_argument("no") is False
+    assert boolean_argument("y") is True
+    assert boolean_argument("n") is False
+    assert boolean_argument("on") is True
+    assert boolean_argument("off") is False
```
Schamper, line 19:
```suggestion
    assert boolean_argument("off") is False
    assert boolean_argument(True) is True
    assert boolean_argument(False) is False
    assert boolean_argument(1) is True
    assert boolean_argument(0) is False
```

Schamper (changes requested):

PR #171, opened 2025-06-18
Add _field_types attribute to GroupedRecord

### LINKED ISSUE #168: Grouped Record
Hello,
I found a bug in GroupedRecord where it does not call `_generate_record_class` like in the equivalent Record class:  
https://github.com/fox-it/flow.record/blob/f4eba70f65e0f9a268f5e1421b06df1612a3acee/flow/record/base.py#L507  
https://github.com/fox-it/flow.record/blob/f4eba70f65e0f9a268f5e1421b06df1612a3acee/flow/record/base.py#L508  
so when trying to run --hash/--resolve in target-query in does not work because it can't find the fields:  
https://github.com/fox-it/flow.record/blob/f4eba70f65e0f9a268f5e1421b06df1612a3acee/flow/record/base.py#L267  
as you can see there is no init to `self.recordType`

B0TAxy:
When using `--hash`, `--resolve`, or both, there's a script called *record_modifier.py* that tries to find the fields and then performs the requested action, adding a new field afterward. However, since the `GroupedRecord` class doesn’t have the `self._field_types` attribute, it just crashes every time a `GroupedRecord` is involved. This code fixes that issue.

Fixes #168

### DIFF
--- a/tests/test_record.py
+++ b/tests/test_record.py
@@ -151,7 +151,7 @@ def test_grouped_record() -> None:
             ("string", "hello"),
         ],
     )
-    excpected_field_types = {
+    expected_field_types = {
         "hello": fieldtypes.string,
         "world": fieldtypes.string,
         "count": fieldtypes.uint32,
@@ -196,9 +196,9 @@ def test_grouped_record() -> None:
     assert rdict["profile"] == "omg"
     assert rdict["count"] == 12345
 
-    # Test gourped._field_types
+    # Test grouped._field_types
     assert grouped._field_types
-    assert grouped._field_types == excpected_field_types
+    assert grouped._field_types == expected_field_types
 
 
 def test_grouped_records_packing(tmp_path: Path) -> None:


B0TAxy:
> Thanks! Could you also add a regression test that tests the old broken situation?

Hi,
Thanks for the review. I've added a regression test — please let me know if it's sufficient.

yunzheng (changes requested):
Looks good, thanks for the pointer where it goes wrong. It seems that dissect.target is using a private attribute to get the fields.

Can you apply the above suggestions by copilot? PR looks good otherwise for merge.

B0TAxy:
> Looks good, thanks for the pointer where it goes wrong. It seems that dissect.target is using a private attribute to get the fields.
> 
> Can you apply the above suggestions by copilot? PR looks good otherwise for merge.

I've addressed the suggestions provided by Copilot.

yunzheng (approved):

PR #172, opened 2025-06-19
Fix Ruff linter violations across the codebase

B0TAxy:
This PR addresses and fixes multiple Ruff linter warnings and errors to improve code quality and consistency across the project. The changes include:

* Adding missing `tz` parameters to `datetime.now()` calls to ensure timezone-aware datetimes (DTZ005).
* Replacing deprecated or discouraged patterns according to the active Ruff ruleset.
* Formatting code to comply with style rules (e.g., unused imports, quote styles, whitespace issues).
* General cleanup to reduce lint noise and prevent potential runtime issues.

No functional changes were made — these updates are strictly linter-related.

### DIFF
--- a/examples/filesystem.py
+++ b/examples/filesystem.py
@@ -32,9 +32,9 @@ FilesystemFile = RecordDescriptor(descriptor)
 
 def hash_file(path: str | Path) -> None:
     with Path(path).open("rb") as f:
-        while 1:
+        while True:
             d = f.read(4096)
-            if d == "":
+            if not d:
                 break
 
 
@@ -48,14 +48,14 @@ class FilesystemIterator:
     def classify(self, source: str, classification: str) -> None:
         self.recordType = FilesystemFile.base(_source=source, _classification=classification)
 
-    def iter(self, path: Path) -> Iterator[FilesystemFile]:
-        return self._iter(path.resolve())
+    def iter(self, path: str | Path) -> Iterator[FilesystemFile]:
+        return self._iter(Path(path).resolve())
 
     def _iter(self, path: Path) -> Iterator[FilesystemFile]:
-        if "proc" in path.parts:
+        if path.is_relative_to("/proc"):
             return
 
-        st = os.lstat(path)
+        st = path.lstat()
 
         abspath = path
         if self.basepath and abspath.startswith(self.basepath):
@@ -75,9 +75,9 @@ class FilesystemIterator:
             size=st.st_size,
             uid=st.st_uid,
             gid=st.st_gid,
-            ctime=datetime.fromtimestamp(st.st_ctime, tz=ZoneInfo("UTC")),
-            mtime=datetime.fromtimestamp(st.st_mtime, tz=ZoneInfo("UTC")),
-            atime=datetime.fromtimestamp(st.st_atime, tz=ZoneInfo("UTC")),
+            ctime=st.st_ctime,
+            mtime=st.st_mtime,
+            atime=st.st_atime,
             link=link,
         )
 
--- a/examples/filesystem.py
+++ b/examples/filesystem.py
@@ -1,9 +1,17 @@
+from __future__ import annotations
+
 import os
 import stat
-
 from datetime import datetime
+from pathlib import Path
+from typing import TYPE_CHECKING
+from zoneinfo import ZoneInfo
 
 from flow.record import RecordDescriptor, RecordWriter
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
 
 descriptor = """
 filesystem/unix/entry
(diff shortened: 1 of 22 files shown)


examples/tcpconn.py
```diff
@@ -1,6 +1,7 @@
 import random
-
 from datetime import datetime
+from zoneinfo import ZoneInfo
```
yunzheng, line 3:
```python
from datetime import timezone
...
UTC = timezone.utc
```

examples/tcpconn.py
```diff
@@ -32,9 +33,9 @@
 
 rs = record.RecordWriter()
 
-for i in range(500):
+for _ in range(500):
     r = conn(
-        ts=datetime.now(),
+        ts=datetime.now(tz=ZoneInfo("UTC")),
```
yunzheng, line 38: reference the toplevel UTC variable.

examples/filesystem.py
```diff
@@ -22,31 +30,29 @@
...
-    def iter(self, path):
-        path = os.path.abspath(path)
-        return self._iter(path)
+    def iter(self, path: Path) -> Iterator[FilesystemFile]:
+        return self._iter(path.resolve())
 
-    def _iter(self, path):
-        if path.startswith("/proc"):
+    def _iter(self, path: Path) -> Iterator[FilesystemFile]:
+        if "proc" in path.parts:
```
yunzheng, line 55: This would also match on `/tmp/proc/bla` instead of only `/proc/*`. Recommended to use:

```suggestion
        if path.is_relative_to("/proc"):
```

examples/filesystem.py
```diff
@@ -22,31 +30,29 @@
...
             return
 
         st = os.lstat(path)
```
yunzheng, line 58:
```suggestion
        st = path.lstat()
```

examples/filesystem.py
```diff
@@ -69,20 +75,19 @@ def _iter(self, path):
             size=st.st_size,
             uid=st.st_uid,
             gid=st.st_gid,
-            ctime=datetime.fromtimestamp(st.st_ctime),
-            mtime=datetime.fromtimestamp(st.st_mtime),
-            atime=datetime.fromtimestamp(st.st_atime),
+            ctime=datetime.fromtimestamp(st.st_ctime, tz=ZoneInfo("UTC")),
+            mtime=datetime.fromtimestamp(st.st_mtime, tz=ZoneInfo("UTC")),
+            atime=datetime.fromtimestamp(st.st_atime, tz=ZoneInfo("UTC")),
```
yunzheng, line 80: we can just pass epoch values directly to the Record initializer now: 

```suggestion
            ctime=st.st_ctime,
            mtime=st.st_mtime,
            atime=st.st_atime,
```

tests/test_fieldtypes.py
```diff
@@ -376,15 +376,15 @@ def test_uri_type() -> None:
...
 
     r = TestRecord()
-    with pytest.warns(DeprecationWarning):
+    with pytest.warns(DeprecationWarning):  # noqa
         r.path = uri.normalize(r"c:\Users\Fox-IT\Downloads\autoruns.exe")
     assert r.path.filename == "autoruns.exe"
-    with pytest.warns(DeprecationWarning):
+    with pytest.warns(DeprecationWarning):  # noqa
         assert r.path.dirname == uri.normalize(r"\Users\Fox-IT\Downloads")
     assert r.path.dirname == "/Users/Fox-IT/Downloads"
```
yunzheng, line 389: Please add the extra `match` regex attribute to `pytest.warns` to match/test the actual warning, so we don't have to do `# noqa`

examples/filesystem.py
```diff
@@ -22,31 +30,29 @@
...
 
-    def classify(self, source, classification):
+    def classify(self, source: str, classification: str) -> None:
         self.recordType = FilesystemFile.base(_source=source, _classification=classification)
 
-    def iter(self, path):
-        path = os.path.abspath(path)
-        return self._iter(path)
+    def iter(self, path: Path) -> Iterator[FilesystemFile]:
+        return self._iter(path.resolve())
```
yunzheng, line 52:
```suggestion
    def iter(self, path: str | Path) -> Iterator[FilesystemFile]:
        return self._iter(Path(path).resolve())
```

yunzheng (changes requested):
Thanks for the PR, the example scripts were long overdue for this.

I added some small suggestions.

examples/filesystem.py
```diff
@@ -69,20 +75,19 @@ def _iter(self, path):
...
+            mtime=datetime.fromtimestamp(st.st_mtime, tz=ZoneInfo("UTC")),
+            atime=datetime.fromtimestamp(st.st_atime, tz=ZoneInfo("UTC")),
             link=link,
         )
 
         if ifmt == stat.S_IFDIR:
-            for i in os.listdir(path):
+            for i in path.iterdir():
                 if i in (".", ".."):
                     continue
```
yunzheng, line 87: the lines can be deleted now as we use Pathlib iterdir(), which doesn't return ".." and "." directories:

examples/filesystem.py
```diff
@@ -1,10 +1,18 @@
+from __future__ import annotations
+
 import os
 import stat
-
 from datetime import datetime
+from pathlib import Path
+from typing import TYPE_CHECKING
+from zoneinfo import ZoneInfo
```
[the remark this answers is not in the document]
B0TAxy: You are most certainly correct

tests/test_fieldtypes.py
B0TAxy: what should I write in the match?

yunzheng: this one works: 

```python
with pytest.warns(DeprecationWarning, match="Do not use class uri\(...\) for filesystem paths, use class path\(...\)"):
```

Also use this at `test_packer::test_uri_packing`, as tests are now failing.

yunzheng (changes requested):
There are still some tests failing when I run `tox -e py3`, could you look into this?

I think mainly due to some imports that have moved to global imports. If that caused the side effect I think for now it would be better to change it back and add a specific `# noqa:` flag to the inline import.

tests/test_fieldtypes.py
B0TAxy: I had to move the imports back inside the functions because placing them at the top-level causes some tests to fail for unknown reasons. While the current version of Ruff doesn't flag this as a violation, newer versions do, so we should revisit this later.

yunzheng (approved):
LGTM

PR #178, opened 2025-08-07
Add record descriptor name to Elastic writer exception message

JSCU-CNI:
This PR improves the exception message thrown when the Elastic adapter attempts to write documents that get rejected by the remote Elasticsearch server.

### DIFF
--- a/flow/record/adapter/elastic.py
+++ b/flow/record/adapter/elastic.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import hashlib
+import json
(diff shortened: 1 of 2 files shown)


JSCU-CNI:
> I was thinking maybe we can use Exception.add_note() here, that could be more clean.

That looks like a clean solution! As long as the message in the test is exposed we're happy.

> Is this something you can explore for this PR? If not, i'm also happy to check.

Due to limited resources currently not unfortunately.

PR #180, opened 2025-08-11
Fix JsonRecordPacker when bytes field is None

### LINKED ISSUE #179: Rdump crashes on fieldtype `bytes` with the value `None`
If a record has a field with the type `bytes` and value of `None`, rdump will fail because it cannot base64 decode `None`.

```zsh
target-query -f notifications -t <TARGET> -j > notifications.json
rdump notifications.json -c 1
  File "flow.record/flow/record/jsonpacker.py", line 102, in unpack_obj
    obj[field_name] = base64.b64decode(obj[field_name])
                      ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
[...]
    raise TypeError("argument should be a bytes-like object or ASCII "
                    "string, not %r" % s.__class__.__name__) from None
TypeError: argument should be a bytes-like object or ASCII string, not 'NoneType'
XXXX [error    ] rdump encountered a fatal error: argument should be a bytes-like object or ASCII string, not 'NoneType' [flow.record.tools.rdump]
```

This occurred while using several Dissect Target plugins like`notifications`, `browser.passwords`, etc.

yunzheng:
Fixes #179

### DIFF
--- a/tests/adapter/test_sqlite_duckdb.py
+++ b/tests/adapter/test_sqlite_duckdb.py
@@ -169,11 +169,12 @@ def test_read_from_sqlite(tmp_path: Path, db: Database) -> None:
             """
         )
         for i in range(1, 30):
+            dt_isoformat = datetime(2023, 10, i, 13, 37, tzinfo=timezone.utc).isoformat()
             con.execute(
                 """
                 INSERT INTO 'test/record' VALUES (?, ?, ?, ?)
                 """,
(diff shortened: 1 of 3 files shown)


tests/adapter/test_sqlite_duckdb.py
Schamper, line 1: This change seems irrelevant?

Schamper (changes requested):

tests/adapter/test_sqlite_duckdb.py
yunzheng: Yes, I noticed a deprecation warning during pytest, I was lazy to create a dedicated PR :D

PR #181, opened 2025-08-13
windows_path __repr__ does not work if it contains surrogate characters

respondersGY:
Closes https://github.com/fox-it/dissect.target/issues/1276

### DIFF
--- a/flow/record/stream.py
+++ b/flow/record/stream.py
@@ -45,7 +45,7 @@ class RecordPrinter:
         self.auto_flush = flush
 
     def write(self, obj: Record) -> None:
-        buf = repr(obj).encode(errors="surrogateescape") + b"\n"
+        buf = repr(obj).encode() + b"\n"
         self.fp.write(buf)
         if self.auto_flush:
             self.flush()


yunzheng:
So it indeed looks more to be an issue with `windows_path()` not able to handle surrogate characters when doing repr. I modified it's repr to this:

```python
class windows_path(pathlib.PureWindowsPath, path):
    def __repr__(self) -> str:
        s = str(self)
        # Only use repr() if we have surrogates that need escaping
        try:
            s.encode('utf-8')
        except UnicodeEncodeError:
            # Has surrogates - use repr but fix the over-escaping
            s = repr(s)[1:-1]  # This escapes surrogates as \udcXX
            s = s.replace('\\\\', '\\')  # Fix double backslashes
            s = s.replace("\\'", "'")   # Fix over-escaped quotes
            s = s.replace('\\"', '"')   # Fix over-escaped double quotes        
    
        quote = "'"
        if "'" in s:
            if '"' in s:
                s = s.replace("'", "\\'")
            else:
                quote = '"'
        return f"{quote}{s}{quote}"
```

Then it seems to work (without the change in the RecordPrinter)

yunzheng (changes requested):
added small remark, looks good otherwise

yunzheng (approved):

PR #183, opened 2025-08-18
Simplify the command type

Miauwkeru:
Instead of changing the behaviour of the adapters to accomodate the command type like in #130
I opt instead to simplify the command type itself and make the structure completely flat.
If we ever want to change how it gets formatted for an adapter, it feels like this should happen at the adapter level
instead of the packing level

### DIFF
--- a/flow/record/fieldtypes/__init__.py
+++ b/flow/record/fieldtypes/__init__.py
@@ -800,7 +800,7 @@ class command(FieldType):
         return False
 
     def _split(self, value: str) -> tuple[str, tuple[str, ...]]:
-        if not (value):
+        if not value:
             return "", ()
 
         executable, *args = shlex.split(value, posix=self.__path_type is posix_path)
@@ -839,12 +839,11 @@ class command(FieldType):
             return
 
         if isinstance(val, str):
-            val = tuple(shlex.split(val, posix=self.__path_type is posix_path))
-
-        if val and self.__path_type is windows_path:
-            val = (" ".join(val),)
-
-        self.__args = (*val,)
+            self.__args = tuple(shlex.split(val, posix=self.__path_type is posix_path))
+        elif isinstance(val, list):
+            self.__args = tuple(val)
+        else:
+            self.__args = val
 
     @property
     def raw(self) -> str:
@@ -854,11 +853,11 @@ class command(FieldType):
             exe = shlex.quote(exe)
 
         result = [exe]
+        # Only quote on posix paths as shlex doesn't remove the quotes on non posix paths
         if self.__path_type is posix_path:
             result.extend(shlex.quote(part) if " " in part else part for part in self.args)
         else:
             result.extend(self.args)
-
         return " ".join(result)
 
     @classmethod
--- a/tests/fieldtypes/test_fieldtypes.py
+++ b/tests/fieldtypes/test_fieldtypes.py
@@ -1065,7 +1065,7 @@ def test_command_record() -> None:
     args = ("-h", "-q", "-something")
     assert isinstance(record.commando.executable, _type)
     assert record.commando.executable == "something.so"
-    assert record.commando.args == (" ".join(args),) if os.name == "nt" else args
+    assert record.commando.args == args
 
 
 def test_command_integration(tmp_path: pathlib.Path) -> None:
@@ -1080,12 +1080,12 @@ def test_command_integration(tmp_path: pathlib.Path) -> None:
         record = TestRecord(commando=r"\.\?\some_command.exe -h,help /d quiet")
         writer.write(record)
         assert record.commando.executable == r"\.\?\some_command.exe"
-        assert record.commando.args == (r"-h,help /d quiet",)
+        assert record.commando.args == (r"-h,help", "/d", "quiet")
 
     with RecordReader(tmp_path / "command_record") as reader:
         for record in reader:
             assert record.commando.executable == r"\.\?\some_command.exe"
-            assert record.commando.args == (r"-h,help /d quiet",)
+            assert record.commando.args == (r"-h,help", "/d", "quiet")
             assert record.commando.raw == r"\?\some_command.exe -h,help /d quiet"
 
 
@@ -1149,9 +1149,9 @@ def test_integration_correct_path(tmp_path: pathlib.Path) -> None:
             ("something,or,somethingelse",),
         ),
         # Test a single quoted path
-        (r"'c:\path to some exe' /d /a", r"c:\path to some exe", (r"/d /a",)),
+        (r"'c:\path to some exe' /d /a", r"c:\path to some exe", ("/d", "/a")),
         # Test a double quoted path
-        (r'"c:\path to some exe" /d /a', r"c:\path to some exe", (r"/d /a",)),
+        (r'"c:\path to some exe" /d /a', r"c:\path to some exe", ("/d", "/a")),
         # Test a unquoted path
         (r"\Users\test\hello.exe", r"\Users\test\hello.exe", ()),
         # Test an unquoted path with a path as argument
@@ -1159,7 +1159,7 @@ def test_integration_correct_path(tmp_path: pathlib.Path) -> None:
         # Test a quoted UNC path
         (r"'\\192.168.1.2\Program Files\hello.exe'", r"\\192.168.1.2\Program Files\hello.exe", ()),
         # Test an unquoted UNC path
-        (r"\\192.168.1.2\Users\test\hello.exe /d /a", r"\\192.168.1.2\Users\test\hello.exe", (r"/d /a",)),
+        (r"\\192.168.1.2\Users\test\hello.exe /d /a", r"\\192.168.1.2\Users\test\hello.exe", ("/d", "/a")),
         # Test an empty command string
         (r"''", r"", ()),
     ],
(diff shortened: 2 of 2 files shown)


tests/adapter/test_json.py
```diff
@@ -1,24 +1,42 @@
...
+    COMMAND = auto()
+
+
+FUNCTIONS = {
+    GeneratorType.RECORDS: generate_records,
+    GeneratorType.COMMAND: generate_command_records,
+}
+
+
+@pytest.mark.parametrize("generator_function", FUNCTIONS.keys())
```
yunzheng, line 30: I can see what you are getting at, but I would prefer a specific test that would then just have fieldtypes parametrized or auto generated.

For now I would just add an extra test in the likes of `test_json_command_fieldtype"

tests/_utils.py
```diff
@@ -42,3 +42,16 @@ def generate_plain_records(count: int = 100) -> Iterator[Record]:
...
+    TestRecord = RecordDescriptor(
+        "test/addapter/command",
+        [
+            ("uint32", "number"),
+            ("command", "data"),
+        ],
+    )
+
+    for i in range(count):
+        yield TestRecord(number=i, data=f"/path/to/file {i}")
```
yunzheng, line 57: I'm not sure if we really need a specific generator for one specific field type. Also, it doesn't generate windows commands.

I think it would be better to just create specific TestRecord in your tests and with some specific real life example commands.

yunzheng (changes requested):
Before I continue reviewing, it doesn't seem to pass these simple tests:

```
def test_empty_command() -> None:
    command = fieldtypes.command()
    assert command.executable == None
    assert command.args == None


def test_empty_string_command() -> None:
    command = fieldtypes.command("")
    assert command.executable == ""
    assert command.args == []
```

Would error with:

```
flow/record/fieldtypes/__init__.py:766: in __init__
    self.executable, self.args = self._split(self._raw)
                                 ^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = (executable=None, args=None), value = ''

    def _split(self, value: str) -> tuple[path, list[str]]:
>       executable, *args = shlex.split(value, posix=self._posix)
        ^^^^^^^^^^^^^^^^^
E       ValueError: not enough values to unpack (expected at least 1, got 0)
```

Also i'm not sure if my asserts would match with what one would expect. That is up for debate.

flow/record/fieldtypes/__init__.py
```diff
@@ -776,30 +772,33 @@ def __eq__(self, other: object) -> bool:
...
-        return shlex.join([str(self.executable), *self.args])
+        return self._raw
 
-    def _pack(self) -> tuple[tuple[str, list], str]:
+    def _pack(self) -> tuple[str, int]:
         command_type = TYPE_WINDOWS if isinstance(self, windows_command) else TYPE_POSIX
         if self.executable:
-            _exec, _ = self.executable._pack()
-            return ((_exec, self.args), command_type)
+            return (self._raw, command_type)
```
yunzheng, line 797: maybe no need to also store the `command_type` to make it even more simpler, as you already deduce from str in `__init__`

tests/adapter/test_json.py
```diff
@@ -117,3 +118,40 @@ def test_json_adapter_with_record_descriptors(tmp_path: Path, record_adapter_pat
...
+    TestRecord = RecordDescriptor(
+        "test/command",
+        [
+            ("command", "commando"),
+        ],
+    )
+
+    writer.write(
+        TestRecord(
+            commando=fieldtypes.windows_command("C:\\help.exe data"),
```
yunzheng, line 137: does initializing with the types test anything specific? why not just command="<string>" here? 

and then also test if it correctly detects it as posix_command or windows_command.

flow/record/fieldtypes/__init__.py
```diff
@@ -736,11 +736,12 @@ class command(FieldType):
     executable: path | None = None
```
yunzheng, line 736: Not specifically touched in the PR, but comment on previous implementation.

Can we not just do `self.executable = path(executable)` and let path field type itself detect if it's windows or posix? would that work?

tests/fieldtypes/test_fieldtypes.py
```diff
@@ -1024,6 +1024,24 @@ def test_datetime_comparisions() -> None:
...
 
+@pytest.mark.parametrize(
+    "command_cls",
+    [
+        fieldtypes.posix_command,
+        fieldtypes.windows_command,
+        fieldtypes.command,
+    ],
+)
+def test_empty_command(command_cls: type[command]) -> None:
```
yunzheng, line 1035: do you know of any other corner cases that we could test?

yunzheng (changes requested):

tests/fieldtypes/test_fieldtypes.py
yunzheng: This can be a real life example from the Windows RunMRU list, mainly concerning spaces in paths:

```python
def test_record_windows_command_spaces() -> None:
    TestRecord = RecordDescriptor(
        "test/command",
        [
            ("command", "value"),
        ],
    )

    r = TestRecord(r"c:\users\John Doe\Desktop\malware.exe /all /the /things")
    assert r.value.executable == r"c:\users\John Doe\Desktop\malware.exe"
    assert r.value.args == ["/all", "/the", "/things"]
```

But it gets parsed as:

`(executable='c:\users\John', args=['Doe\\Desktop\\malware.exe /all /the /things'])`

This is a tricky one though, and i'm not sure what a good solution would be.

It's possible that if one decides to use the `command` fieldtype, needs to do some parsing/tweaking the string beforehand before initializing the `command` field.

yunzheng (changes requested):

flow/record/fieldtypes/__init__.py
```diff
@@ -795,66 +793,78 @@ def __eq__(self, other: object) -> bool:
...
+    @property
+    def args(self) -> tuple[str, ...]:
+        return self.__args
 
-class posix_command(command):
-    _posix = True
-    _path_type = posix_path
+    @executable.setter
+    def executable(self, val: str | path | None) -> None:
+        self.__executable = self.__path_type(val)
```
yunzheng: with the latest merged PR #200 we should be able to directly initialize path with the value to determine the correct path type. Then you can als get rid of `__path_type`.

tests/fieldtypes/test_fieldtypes.py
```diff
@@ -1041,15 +1053,19 @@ def test_command_record() -> None:
...
-    assert record.commando.args == ["-h"]
+    assert record.commando.args == ("-h",)
 
     record = TestRecord(commando="something.so -h -q -something")
-    assert isinstance(record.commando, posix_command)
+    args = ("-h", "-q", "-something")
+    assert isinstance(record.commando.executable, _type)
     assert record.commando.executable == "something.so"
-    assert record.commando.args == ["-h", "-q", "-something"]
+    assert record.commando.args == (" ".join(args),) if os.name == "nt" else args
```
yunzheng: can you explain why you do the os.name check here?

Miauwkeru: because of https://github.com/fox-it/flow.record/blob/cd69cf171af3d92385facb33ad8c2e8593a75e3a/flow/record/fieldtypes/__init__.py#L634

It will chose to be a windows or linux path type depending on the os it runs on. And on windows, arguments are interpreted differently depending on the executable itself, so in that case the arguments are joined together in one single string.

flow/record/fieldtypes/__init__.py
Miauwkeru: Do you mean removing the whole `__path_type` in general? As in that case, there will be issues with for example `hello_world.exe` that will be interpreted differently depending on what system it gets run on.

If you just mean this specific case, this could still be an issue because the whole type can change from windows to posix path and vise versa. Tho I already have an idea on how to fix that specific issue.

yunzheng: ah wait I thought __path_type was a function to determine the path type.

flow/record/fieldtypes/__init__.py
```diff
@@ -795,66 +793,78 @@ def __eq__(self, other: object) -> bool:
...
+            return
 
-class posix_command(command):
-    _posix = True
-    _path_type = posix_path
+        if isinstance(val, str):
+            val = tuple(shlex.split(val, posix=self.__path_type is posix_path))
 
+        if val and self.__path_type is windows_path:
+            val = (" ".join(val),)
```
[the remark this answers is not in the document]
Schamper: For documentation sake, I believe this was originally the choice because _technically_ that's correct on Windows. IIRC, in Linux the kernel passes split arguments to your new process (and if you start a new process, you give it an array of arguments). On Windows, the program is responsible for splitting/parsing the arguments, the kernel will pass it as one big string to you (and if you start a new process, you give it one big string of all your arguments).

Resources:
- https://daviddeley.com/autohotkey/parameters/parameters.htm#WINCREATE
- https://nullprogram.com/blog/2022/02/18/
  - https://github.com/python/cpython/blob/3.10/Lib/subprocess.py#L529
  - https://go.googlesource.com/go/+/refs/tags/go1.17.7/src/syscall/exec_windows.go#101

For what it's worth, I do agree that split is a much more sane representation.

PR #184, opened 2025-08-18
Improve progress bar by tracking record metrics

yunzheng:
All record reader adapters now utilize an `AppContext` to track metrics such as the number of records read, matched, and excluded.

This context is accessible via `flow.record.context.get_app_context()`.

The progress bar in `rdump` has been updated to display these metrics.


flow/record/context.py
```diff
@@ -0,0 +1,53 @@
...
+import sys
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+APP_CONTEXT: ContextVar[AppContext] = ContextVar("app_context")
```
Schamper, line 12:
```suggestion
APP_CONTEXT: ContextVar[AppContext] = ContextVar("APP_CONTEXT")
```

flow/record/context.py
```diff
@@ -0,0 +1,53 @@
...
+    app_dataclass = dataclass
+
+
+@app_dataclass
+class AppContext:
+    """Context for the application, holding metrics like amount of processed records."""
+
+    records_read: int = 0
+    records_matched: int = 0
+    records_excluded: int = 0
```
Schamper, line 51: Could the names be simplified to `read`, `matched`, `excluded`, since it's fairly obvious it's to do with records in this context (🥁)?

flow/record/adapter/avro.py
```diff
@@ -121,8 +123,12 @@ def __iter__(self) -> Iterator[record.Record]:
                     obj[field_name] = EPOCH + timedelta(microseconds=value)
 
             rec = self.desc.recordType(**obj)
+            ctx.records_read += 1
             if not self.selector or self.selector.match(rec):
```
Schamper, line 127: Maybe an idea that you can "attach" the context to the selector, and let the selector increment `records_matched` and `records_excluded`? Easiest way would be to just add a second argument.

flow/record/stream.py
```diff
@@ -137,8 +140,12 @@ def __iter__(self) -> Iterator[Record]:
                 if isinstance(obj, RecordDescriptor):
                     self.packer.register(obj)
                 else:
-                    if not self.selector or self.selector.match(obj):
+                    ctx.records_read += 1
+                    if not selector_match or selector_match(obj):
```
Schamper, line 144: Maybe do this pattern in the adapters too?

Schamper (changes requested):

flow/record/adapter/avro.py
yunzheng: I opted for a helper function called `match_record_with_context` in the selector namespace.

One could argue it can also increase `ctx.read`, then you don't have to do any metrics tracking in the adapter itself besides just calling this function on every record.

yunzheng:
Also still contemplating if I should change `ctx.excluded` to `ctx.unmatched`. As during development I kept forgetting what the exact attribute name was.

flow/record/adapter/avro.py
Schamper: > One could argue it can also increase `ctx.read`, then you don't have to do any metrics tracking in the adapter itself besides just calling this function on every record.

Hm if you do that, I'd maybe move the helper function to `flow.record.context` instead.

flow/record/context.py
yunzheng: i'll just keep it AppContext.

flow/record/tools/rdump.py
```diff
@@ -77,6 +78,49 @@ def list_adapters() -> None:
...
+
+        def update_progress_bar(self) -> None:
+            source_count = self.ctx.source_count
+            source_total = self.ctx.source_total
+            read = self.ctx.records_read
+            matched = self.ctx.records_matched
+            excluded = self.ctx.records_excluded
+
+            self.progress_bar.n = read
+            postfix = f"source={source_count}/{source_total}, {read=}, {matched=}, {excluded=}"
```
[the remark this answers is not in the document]
yunzheng: done, thanks CoPilot

flow/record/adapter/avro.py
```diff
@@ -113,6 +114,8 @@ def __init__(self, path: str, selector: str | None = None, **kwargs):
         }
 
     def __iter__(self) -> Iterator[record.Record]:
+        ctx = get_app_context()
+        selector = self.selector
```
Schamper, line 118: The setup for this pattern looks a little awkward, but I recognize it's for performance reasons, so I think it should be fine to keep it like this.

flow/record/adapter/sqlite.py
```diff
@@ -195,13 +196,14 @@ def read_table(self, table_name: str) -> Iterator[Record]:
...
+        ctx = get_app_context()
+        selector = self.selector
         for table_name in self.table_names():
             self.logger.debug("Reading table: %s", table_name)
             for record in self.read_table(table_name):
-                if not self.selector or self.selector.match(record):
+                if match_record_with_context(record, selector, ctx):
                     yield record
 
-
```
Schamper, line 204: Won't this fail linting? Thought you needed two empty whitespaces between classes.

flow/record/context.py
```diff
@@ -0,0 +1,78 @@
...
+    from flow.record.selector import Selector
+
+APP_CONTEXT: ContextVar[AppContext] = ContextVar("APP_CONTEXT")
+
+
+def get_app_context() -> AppContext:
+    """Retrieve the application context, creating it if it does not exist.
+
+    Returns:
+        AppContext: The application context.
```
Schamper, line 22:
```suggestion
        The application context.
```
Return type is already provided in the type hint.

Schamper (changes requested):

flow/record/adapter/sqlite.py
Schamper: There's a `ruff format --check flow tests` missing in the `tox.ini` config for linting.

PR #188, opened 2025-08-20
Fix BrokenPipeError handling in rdump

### LINKED ISSUE #186: @catchsig_pipe not working in rdump
It seems that the `@catch_sigpipe` decorator stopped working since this change in rdump that catches any exceptions in the finally block during __exit__:

```python
    finally:
        if record_writer:
            # Exceptions raised in threads can be thrown when deconstructing the writer.
            try:
                record_writer.__exit__()
            except Exception as e:
                print_error(e)
                ret = 1
```

This should be investigated if `RecordWriter.__exit__()` needs to raise any exceptions at all.

yunzheng:
Fixes #186


yunzheng:
@JSCU-CNI can you also check if this PR works for you regarding Elastic search exceptions? as I basically reverted the RecordWriter back to a `with` statement that was removed in this commit: c482853a26f7ce541505266bfa325c5aa187fb91

tests/tools/test_rdump.py
```diff
@@ -720,3 +720,39 @@ def test_rdump_list_progress(tmp_path: Path, capsys: pytest.CaptureFixture) -> N
 
     # stdout should contain the RecordDescriptor definition and count
     assert "# <RecordDescriptor test/rdump/progress, hash=eeb21156>" in captured.out
+
+
+def test_rdump_catch_sigpipe(tmp_path: Path) -> None:
+    """Test if rdump properly suppresses BrokenPipeError when writing to a closed file handle."""
+    if platform.system() == "Windows":
+        pytest.skip("skipping this test on windows")
```
Schamper, line 728:
```suggestion
@pytest.mark.skipif(platform.system() == "Windows", "skipping this test on Windows")
def test_rdump_catch_sigpipe(tmp_path: Path) -> None:
    """Test if rdump properly suppresses BrokenPipeError when writing to a closed file handle."""
```

Schamper (changes requested):

PR #189, opened 2025-08-20
Fix version pinning for zstandard on pypy3.9

### LINKED ISSUE #185: Installing zstandard 0.24.0 breaks on windows+pypy3.9
`zstandard==0.24.0` was released yesterday (17 August 2025) and it starts failing our CI for: `windows-latest, pypy3.9`:

```
  tmprieep2zh.h

      <string>:62: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.

      <string>:63: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.

      Traceback (most recent call last):

        File "C:\hostedtoolcache\windows\PyPy\3.9.19\x86\Lib\cffi\_shimmed_dist_utils.py", line 33, in <module>

          from distutils.msvc9compiler import MSVCCompiler

      ModuleNotFoundError: No module named 'distutils.msvc9compiler'
```

Because it tries to compile it means there is probably also no wheel for this.

yunzheng:
~~Ordering matters to ensure that the pinned version takes precedence~~

Fixes #185

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ repository = "https://github.com/fox-it/flow.record"
 compression = [
     "lz4",
     "zstandard==0.23.0; platform_python_implementation == 'PyPy' and python_version == '3.9'", # Pin to last working for PyPy3.9
-    "zstandard",
+    "zstandard; platform_python_implementation != 'PyPy' and python_version != '3.9'", # Otherwise, pick the latest
 ]
 elastic = [
     "elasticsearch",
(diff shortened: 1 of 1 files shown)


yunzheng:
Looks to be an issue with tox, it doesn't respect the order and will sort it before passing to pip like:

```
pypy3: 3861 W install_package_deps> python -I -m pip install lz4 'msgpack>=0.5.2' 'tzdata; platform_system == "Windows"' zstandard 'zstandard==0.23.0; platform_python_implementation == "PyPy" and python_version == "3.9"' [tox/tox_env/api.py:461]
```

If the bare `zstandard` comes after the one with the constraint, it would work.

PR #205, opened 2026-01-19
Fix JsonRecordPacker for boolean fieldtype

### LINKED ISSUE #204: Boolean packed incorrectly by JSON packer
`flow.record.fieldtypes.boolean` does not identify itself as a boolean value, which results in incorrect JSON encoding in the JSON packer (and by proxy the Elastic adapter).

```python
>>> from flow.record.fieldtypes import boolean
>>> foo = boolean(1)
>>> isinstance(foo, bool)
False
>>> isinstance(foo, int)
True
```

Perhaps this can be fixed with a metaclass `__instancecheck__` method?

JSCU-CNI:
Attempts to fix #204

### DIFF
--- a/tests/fieldtypes/test_boolean.py
+++ b/tests/fieldtypes/test_boolean.py
@@ -29,7 +29,7 @@ def test_boolean() -> None:
     assert repr(r.boolfalse) == "False"
 
     with pytest.raises(ValueError, match="Value not a valid boolean value"):
-        r = TestRecord(2, -1)
+        TestRecord(2, -1)
 
     with pytest.raises(ValueError, match="invalid literal for int"):
-        r = TestRecord("True", "False")
+        TestRecord("True", "False")


JSCU-CNI:
Thank you for taking the time to look into this.

> I went a bit in the JSONEncoder rabbit hole, so, json.dumps() has a cls= and a default= argument. We currently only use default= for unknown types.
> Because fieldtypes.boolean inherits from int (because we cannot inherit from bool), it's a basic known type and the default=x will not be called on this field, as it's only called on "unknown" types that JSON cannot serialize.
> 
> If we want to change the serialization of basic types, we would need our own JSONEncoder class where we override encode() and iterencode() methods, but that seems like a bigger refactor if we want to go that route.

Yes I reached the same conclusion there, perhaps a custom JSONEncoder in the future would be better suited, to prevent duplicate serialization logic in the json packer for (ordered)dicts and records.

> I therefore opted in to just handle an extra conversion step before passing it to json.dumps() that enforces the conversion of fieldtypes.boolean to an actual bool.

This seems like a neat solution for now.

PR #216, opened 2026-02-23
Fix SQLite adapter with GroupedRecords

### LINKED ISSUE #215: Grouped records fail for SQLite adapter
Trying to convert record output from `walkfs` to SQLite database with `rdump`  fails if `--hash` is passed to `target-query`.

Running `target-query -f walkfs --hash some.target | rdump -w sqlite://some.target.records.db` fails. 

`walkfs` with the `--hash` parameter yields grouped records (https://github.com/fox-it/dissect.target/blob/18caf974b6b8ec6b57b4c5dc28be6fb3bf3a4397/dissect/target/helpers/record_modifier.py#L123).

When trying to convert such records to an SQLite database with `rdump`, it fails. This seems to be because the SQLite adapter inserts values based on a record's `__slots__` attribute. For grouped records this attribute is empty and results in an empty tuple being passed to  `prepare_insert_sql()`:
https://github.com/fox-it/flow.record/blob/e9118cf22fedd6059945781bd2e39d965ff73faa/flow/record/adapter/sqlite.py#L115
Which means that `value_placeholder ` is empty here:
https://github.com/fox-it/flow.record/blob/e9118cf22fedd6059945781bd2e39d965ff73faa/flow/record/adapter/sqlite.py#L106
And the SQL below fails:
https://github.com/fox-it/flow.record/blob/e9118cf22fedd6059945781bd2e39d965ff73faa/flow/record/adapter/sqlite.py#L107

Example of a failed run:

```bash
$ uv run target-query -q -f walkfs --hash target.vmx | uv run rdump -w sqlite://./walkfs.records.db
[reading from stdin]
2026-02-23T19:29:04.095644Z [error    ] rdump encountered a fatal error: near ")": syntax error [flow.record.tools.rdump]
```

yunzheng:
This PR ensures that we don't use `record.__slots__` but rather `rdict.keys()` for creating the sqlite table. Small regression test is included.

Also fixes some other small issues I encountered with GroupedRecord class, mainly by renaming the class variables to dunder variables.

fixes #215

### DIFF
--- a/flow/record/adapter/sqlite.py
+++ b/flow/record/adapter/sqlite.py
@@ -109,8 +109,8 @@ def prepare_insert_sql(table_name: str, field_names: tuple[str]) -> str:
 
 def db_insert_record(con: sqlite3.Connection, record: Record) -> None:
     """Insert a record into the database."""
-    desc = record._desc
-    table_name = desc.name
+    descriptor = record._desc
+    table_name = descriptor.name
     rdict = record._asdict()
     sql = prepare_insert_sql(table_name, tuple(rdict.keys()))
 


flow/record/adapter/sqlite.py
```diff
@@ -109,10 +109,10 @@ def prepare_insert_sql(table_name: str, field_names: tuple[str]) -> str:
 
 def db_insert_record(con: sqlite3.Connection, record: Record) -> None:
     """Insert a record into the database."""
-    table_name = record._desc.name
+    desc = record._desc
```
skepppy: Maybe rewrite as `descriptor`, I thought it meant description

skepppy (commented):
Minor comment, but looks good to me otherwise

flow/record/adapter/sqlite.py
yunzheng: I did change `desc` to `descriptor` in this local variable here though

skepppy (approved):

PR #219, opened 2026-03-26
Preserve metadata for multi timestamp records

### LINKED ISSUE #218: Using multi-timestamp causes loss of _source and other metadata fields
## Description

When using `rdump --multi-timestamp`, metadata fields (`_source`, `_classification`, `_generated`) are set to null or modified on all output records.

Disclosure: I used Claude to help explore and diagnose this bug.

## To Replicate

Run `rdump` and `rdump --multi-timestamp --record-classification TEST` and observe outputs. The `_source` and `_classification` will appear as `null` on the second.

Example Output (Truncated)

`target-query /some/image -f userassist | rdump --record-classification TEST -w jsonfile://userassist.jsonl`

```jsonl
{"hostname": "DESKTOP-KLOQJ0V", "ts": "2017-01-27T00:33:27.668032+00:00", "_source": "/home/user/test-images/HD1.E01", "_classification": "TEST", "_generated": "2026-03-26T19:05:01.081233+00:00", "_version": 1, "_type": "record", "_recorddescriptor": ["windows/registry/userassist", 2720671085]}
```

`target-query /some/image -f userassist | rdump --multi-timestamp --record-classification TEST -w jsonfile://userassist.jsonl`

```jsonl
{"ts": "2017-01-27T00:33:27.668032+00:00", "ts_description": "ts", "hostname": "DESKTOP-KLOQJ0V", "_source": null, "_classification": null, "_generated": "2026-03-26T19:08:18.288666+00:00", "_version": 1, "_type": "record", "_recorddescriptor": ["windows/registry/userassist", 1909757769]}
```

## Bug

Looking into the problem, it appears the issue lies here:

https://github.com/fox-it/flow.record/blob/a594c193b1050b361b48f76d502385f22bcbbbac/flow/record/base.py#L1097-L1129

`iter_timestamped_records()` creates a new `TimestampRecord` and merges it with the original record using `extend_record()` on line 1126.

Because `TimestampRecord` is a newly created `RecordDescriptor`, its metadata fields default to `None`. During the extend process, it returns the first value it finds for each key. Because the new `TimestampRecord` is provided first, the `None` values are used over the original record's real metadata.

This also probably has the result of generating a new `_generated` timestamp, but this might be appropriate given it is when that specific record is generated. However, because these are derived records from an original record, I would personally like for them to all have the original `_generated` timestamp too. I'm unsure what the desired behavior is though.

There is also a small scoping issue with `record` being reused each iteration. `record` is already used as a parameter, so to avoid `record` being inadvertently extended  with each subsequent loop of `fields in dt_fields`, the extended record should be assigned to a new variable.

## Suggested Fix

Explicitly mapping the metadata fields after creation of the `TimestampRecord` seems like the most straightforward approach. It would have to be modified if any metadata fields are added in the future, but does allow more control if, for example, the `_generated` field should not be carried over from the original record.

Additionally, tests should be created that check if the desired metadata fields are preserved.

```python
    record_name = record._desc.name
    for field in dt_fields:
        ts_record = TimestampRecord(getattr(record, field.name), field.name)
1. Preserve metadata from the original record so it isn't shadowed by
1. the newly-created TimestampRecord's None defaults.
           ts_record._source = record._source
           ts_record._classification = record._classification
           ts_record._generated = record._generated
1. we extend ``ts_record`` with original ``record`` so TSRecord info goes first.
           result = extend_record(ts_record, [record], name=record_name)
           yield result
   ```

Jacob-gr:
Preserve metadata for multi timestamp records

Fixes metadata fields (`_source`, `_classification`, `_generated`) being set to null 
when `iter_timestamped_records()` expands a record into multi-timestamped records
by explicitly copying metadata fields to new `TimestampRecord` object.

Also fixes an issue with shadowed `record` variable in loop: use a separate result variable 
instead of reassigning `record` in the loop, which previously caused subsequent iterations 
to extend an already-extended record.

Includes a test to validate metadata values are retained. Due to using a new variable in the loop,
a change to  `test_multi_timestamp_descriptor_cache()` is required to adjust the expected cache misses
and hits. Because we are not mutating `record` each loop, we should only expect 1 cache miss.

fixes #218

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -1121,12 +1121,12 @@ def iter_timestamped_records(record: Record) -> Iterator[Record]:
     # yield a new record for each ``datetime`` field assigned as ``ts``.
     record_name = record._desc.name
     for field in dt_fields:
-        ts_record = TimestampRecord(getattr(record, field.name), field.name)
-        # Preserve metadata from the original record so it isn't shadowed by
-        # the freshly-created TimestampRecord's None defaults in ChainMap.
-        ts_record._source = record._source
-        ts_record._classification = record._classification
-        ts_record._generated = record._generated
+        ts_record = TimestampRecord(
+            ts=getattr(record, field.name),
+            ts_description=field.name,
+            _source=record._source,
+            _classification=record._classification,
+            _generated=record._generated,
+        )
         # we extend ``ts_record`` with original ``record`` so TSRecord info goes first.
-        result = extend_record(ts_record, [record], name=record_name)
-        yield result
+        yield extend_record(ts_record, [record], name=record_name)
--- a/tests/record/test_multi_timestamp.py
+++ b/tests/record/test_multi_timestamp.py
@@ -162,8 +162,8 @@ def test_multi_timestamp_preserves_metadata() -> None:
     )
 
     test_record = TestRecord(
-        ctime=datetime(2020, 1, 1, 1, 1, 1),  # noqa: DTZ001
-        atime=datetime(2022, 11, 22, 13, 37, 37),  # noqa: DTZ001
+        ctime=datetime(2020, 1, 1, 1, 1, 1, tzinfo=timezone.utc),
+        atime=datetime(2022, 11, 22, 13, 37, 37, tzinfo=timezone.utc),
         data="test",
     )
     test_record._source = "/evidence/disk.E01"


Jacob-gr:
@DissectBot agree

Issue #220, opened 2026-03-28
Preserving original unsplit value for the command fieldtype


Jacob-gr:
Background: I am working on an Elastic Integration to assist with ingesting Dissect data into Elastic via Elastic Agent. (The Elastic Ingest pipelines would also allow improved parsing of documents being written via rdump's elasticsearch writer).  I've been going over the dissect output of various functions and figuring out how it could best be mapped to ECS and other helpful fields for analysts. During this process, I've come across a few things that I'd like to bring up for discussion and determine if I am possible missing a flag for a command or if there would be a need for a feature request.

I'm opening an issue regarding a proposal to add a feature to the command fieldtype: an original property that preserves the exact input string before any splitting or normalization takes place. If there is already something like this and I'm just not seeing it, please let me know.

## Observed Problem

While testing the `runkeys` function, I noticed the command executable and args getting incorrectly split due to (what I think) are missing quotes in the original value. This results in a structure like:

```json
"command": {"executable": "%ProgramFiles%\\Windows", "args": ["Mail\\wab.exe /Upgrade"]}
```
When one would expect

```json
"command": {"executable": "%ProgramFiles%\\Windows Mail\\wab.exe", "args": ["/Upgrade"]}
```

Looking into it, I see that this is a recognized issue in the docstring of the `command` FieldType:

https://github.com/fox-it/flow.record/blob/a594c193b1050b361b48f76d502385f22bcbbbac/flow/record/fieldtypes/__init__.py#L755-L870

## Proposed Solution

I could join the the executable string and the args array to get an approximation of the original value; however, I feel there would be benefit in having access to the original, unmodified/unsplit/unstriped value (especially when dealing with forensics). I've read about various whitespace padding and null injection techniques to help store payloads in registry values.

Would it be reasonable to include something like an `executable.raw` or `executable.original` field to provide the original string? I would think it would look something like this on the Dissect JSON output:

```json
"command": {"executable": "%ProgramFiles%\\Windows", "args": ["Mail\\wab.exe /Upgrade"], "original": "%ProgramFiles%\\Windows Mail\\wab.exe /Upgrade"}
```

Or it may be better to just provide as a seperate field all together?

```json
"raw_command": "%ProgramFiles%\\Windows Mail\\wab.exe /Upgrade"
```
[...]

PR #224, opened 2026-04-02
Update docstrings for doctests

Miauwkeru:
Noticed from #220 that the docstring of the command type didn't
represent the correct output anymore since we now also split the windows
arguments.

### DIFF
--- a/flow/record/base.py
+++ b/flow/record/base.py
@@ -1041,7 +1041,7 @@ def normalize_fieldname(field_name: str) -> str:
     This normalizes the name so it can still be used in flow.record.
     Reserved field_names are not normalized.
 
-    .. code-block:: text
+    Example::
 
         >>> normalize_fieldname("my-variable-name-with-dashes")
         'my_variable_name_with_dashes'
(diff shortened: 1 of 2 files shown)


yunzheng (commented):
Actually the code block doesn't really make sense to me. The old one didn't really either now that I look back at it. 

```
'c:\\windows\\malware.exe /info'                      ->   windows_path('c:\\windows\\malware.exe) ('/info',)
'/usr/bin/env bash'                                   ->   posix_path('/usr/bin/env') ('bash',)

# In this situation, the executable path needs to be quoted.
'c:\\user\\John Doe\\malware.exe /all /the /things'   ->   windows_path('c:\\user\\John')                                                                  ('Doe\\malware.exe, '/all', '/the', /things')
```

wouldn't it be better to show how it works with a inline doctest?

Miauwkeru:
> wouldn't it be better to show how it works with a inline doctest?

I changed it to use a doctest

yunzheng (changes requested):

yunzheng (approved):

PR #226, opened 2026-05-12
Add object hook method to JsonfileReader

JSCU-CNI:
This PR adds the method `obj_hook` to the `JsonfileReader` adapter. This allows implementations of `JsonfileReader` to modify object output without redefining the entire `__iter__` method.


yunzheng:
I don't see a real problem with adding this, but was curious what the use-case was besides the one shown in the unit test.

PR #230, opened 2026-06-16
Fix Elastic adapter when attempting to generate notes for non-ApiError exceptions

JSCU-CNI:


flow/record/adapter/elastic.py
```diff
@@ -283,6 +283,10 @@ def create_elasticsearch_error_notes(errors: list[dict] | dict, max_notes: int =
 
     notes = []
     for idx, error in enumerate(errors, 1):
+        # The stack of errors could include different exceptions, such as TransportErrors.
+        if not hasattr(error, "get"):
+            continue
+
```
yunzheng: Hmm it looks like `create_elasticsearch_error_notes` only considered `BulkIndexError`, where `.errors` is a list of Dicts: 

https://github.com/elastic/elasticsearch-py/blob/6df97e3c21adffeb33d0a4c64dc784e618d85f41/elasticsearch/helpers/errors.py#L21

And it seems `TransportError` also has `.errors` but is a tuple of Exception:

https://github.com/elastic/elastic-transport-python/blob/08a5d72e93d6306579b732ef5e2e5a4948134898/elastic_transport/_exceptions.py#L31

I think instead of skipping we should consider it being a dictionary or an Exception (and update the typing), then in the loop we can still add the error as a note:

```python
    notes = []
    for idx, error in enumerate(errors, 1):
        if isinstance(error, Exception):
            # Create formatted Exception note
            note_parts = [
                f"Error {idx}, {error}",
            ]
        elif isinstance(error, dict):
            # Extract index information
            ....
        notes.append("\n".join(note_parts) + "\n")
```

yunzheng (changes requested):
Hmm it looks like `create_elasticsearch_error_notes` only considered `BulkIndexError`, where `.errors` is a list of Dicts: 

https://github.com/elastic/elasticsearch-py/blob/6df97e3c21adffeb33d0a4c64dc784e618d85f41/elasticsearch/helpers/errors.py#L21

And it seems `TransportError` also has `.errors` but is a tuple of Exception:

https://github.com/elastic/elastic-transport-python/blob/08a5d72e93d6306579b732ef5e2e5a4948134898/elastic_transport/_exceptions.py#L31

flow/record/adapter/elastic.py
JSCU-CNI: That makes sense. I opted for a slightly simpler `if/else` clause in [670fc41](https://github.com/fox-it/flow.record/pull/230/commits/670fc41f866948b54315936e28ad948f21014bc1).

